PR #1, opened 2021-11-17
Add protobuf definition; convert server and client to protobuf

jtbandes:


proto/protocol_v1.proto
```diff
@@ -0,0 +1,71 @@
...
+    uint32 id = 1;
+    uint32 channel_id = 2;
+  }
+}
+
+message Unsubscribe {
+  repeated uint32 subscription_ids = 1;
+}
+
+message MessageData {
```
amacneil, line 67:
```suggestion
message Message {
```

proto/protocol_v1.proto
```diff
@@ -0,0 +1,71 @@
+syntax = "proto3";
+
+package foxglove.websocket.v1;
```
amacneil, line 3: how do we handle versioning of the protocol? 

I wonder whether it's worth defining a couple of "version independent" messages, so the client could first send an initial message with `accepted_versions uint8[]` and the server responds with `version uint8` to confirm the protocol version, then every message after that would be from the appropriate major version .proto file. any backwards compatible changes we can just append to this existing proto file, but if we need to make a backwards incompatible change then the client could support both versions and the server can decide which version it wants to speak.

or something like that? not sure if you have something in place already, just brainstorming.

proto/protocol_v1.proto
jtbandes: I don't love MessageData, but `message Message` feels like it will get a bit insane. maybe we borrow the MessageEvent terminology from studio here @defunctzombie ?

proto/protocol_v1.proto
jtbandes: that sounds basically the same as the way we use websocket subprotocols, the advantage of subprotocols is that we don't have to add any client/server logic and it's handled before the connection is fully opened. Search the PR for "foxglove.websocket.v1" and you'll see where those subprotocols are used

amacneil: Oh shit, I never heard of subprotocols before. That's super cool. 

Ok, no need to reinvent the wheel then. Carry on.

jtbandes:
The challenges with avoiding extra copies during de/serialization make me think it's not worth it to use protobuf. The JSON + binary approach currently on the main branch seems like one of the most reasonable solutions.

amacneil:
RIP 🪦

PR #3, opened 2021-11-18
Add spec doc, update README

jtbandes:
Import spec from google doc, along with some updates and reformatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,38 @@
-## Virtualenv usage
+# Foxglove Studio WebSocket protocol libraries
+
+This repository provides a protocol specification and reference implementations enabling [Foxglove Studio](https://github.com/foxglove/studio) to ingest arbitrary “live” streamed data.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,19 +1,38 @@
-## Virtualenv usage
+# Foxglove Studio WebSocket protocol libraries
+
+This repository provides a protocol specification and reference implementations enabling [Foxglove Studio](https://github.com/foxglove/studio) to ingest arbitrary “live” streamed data.
+
+The protocol is encoding-agnostic, i.e. it can support Protobuf messages, ROS 1 or 2 messages, etc. (as long as the desired encoding is supported by both client and server).
```
amacneil, line 5: might be worth calling out that while the protocol is encoding-agnostic, Foxglove Studio only supports a specific set of encodings

jtbandes: that's kinda what I was going for with the parenthetical but maybe it's not clear....I guess in the future this can point to a page on foxglove.dev/docs

amacneil: fair. yeah we can link to official message docs once they exist

PR #4, opened 2021-11-18
Add C++ skeleton

jhurliman:
A skeleton of a C++ websocket server. This is (currently) a concrete implementation that compiles to an executable, not a library meant for reuse in other applications. In that vein, it includes ROS1 serialization code so it can publish topics that Studio will be able to understand both encoding-wise and semantically.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1,0 +1,62 @@
+cmake_minimum_required(VERSION 3.8)
+project(foxsocketpp LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED on)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+option(WERROR "Add -Werror flag to build (turns warnings into errors)" ON)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
+
+if(CONAN_EXPORTED) # in conan local cache
+  include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+  conan_basic_setup()
+else() # in user space
+  include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/conan.cmake)
+  conan_cmake_run(CONANFILE conanfile.py
+                  BASIC_SETUP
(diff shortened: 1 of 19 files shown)


cpp/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+cmake_minimum_required(VERSION 3.8)
+project(foxsocketpp LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED on)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+option(WERROR "Add -Werror flag to build (turns warnings into errors)" ON)
```
jtbandes, line 7: Remove and just make this always on? :)

cpp/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
```
jtbandes, line 2: Remove, we have one at the root

cpp/Makefile
```diff
@@ -0,0 +1,28 @@
...
+clean:
+	rm -rf ./$(BUILD_DIR)
+	# remove remains from running 'make coverage'
+	rm -f *.profraw
+	rm -f *.profdata
+
+format:
+	./scripts/format.sh
+
+.PHONY: test
```
jtbandes, line 28: All these should be phony.
Years ago I saw a great life hack: put `.PHONY: x` right before the `x:` rule so you’ll never forget to add it

cpp/conanfile.py
```diff
@@ -0,0 +1,38 @@
+from conans import ConanFile, CMake, tools
+
+PROJECT = "foxsocketpp"
+
+
+class FoxSocketPpConan(ConanFile):
+    name = PROJECT
+    version = "1.0.0"
+    url = f"https://github.com/jhurliman/{PROJECT}"
```
jtbandes, line 9: Needs updating

cpp/CMakeLists.txt
jhurliman: I've seen this used (and used it myself) in various projects because compilers tend to add more warnings over time. So code that compiles fine with today's llvm 13 may throw warnings with tomorrow's llvm 14. This is a quick unblocker for people who want to use newer/different compilers or different compiler flags without modifying the codebase.

PR #14, opened 2021-12-08
Publish Python package to PyPI

jtbandes:
Published as `foxglove-websocket==0.0.1`


amacneil:
I know zero about pypi so deferring to @defunctzombie here

PR #25, opened 2021-12-15
Fix URL encoding of ds.url param

amacneil:


amacneil:
@jtbandes after this can we publish a new release? I gave a demo today and the incorrect wss: URL tripped me up

jtbandes:
I don't understand this change. The url works fine for me.

jtbandes:
The `wss` vs `ws` change was made a while ago and you might have a cached version of the package. you can use `npx @foxgove/...@latest ...` to always run the latest version

amacneil:
Apparently percent encoding characters is not mandatory. It looks weird to me having `://` inside URL GET params, but I don't feel strongly about it.

PR #26, opened 2021-12-15
Add `@latest` to all npx commands

jtbandes:
Ensures the user always gets the latest version of the example scripts. By default npx might use an old version that's cached.


defunctzombie:
👍

Issue #27, opened 2021-12-15
Example Python server using Protobuf


jtbandes:
We should add an example server that sends Protobuf messages.

PR #24, opened 2021-12-14
Reorganize READMEs and make each package's directions clearer

esthersweon:
**Public-Facing Changes**
- Reorganize READMEs and make each package's directions clearer

PR #29, opened 2021-12-17
Add Protobuf example server to Python package

jtbandes:
**Public-Facing Changes**
- Adds an example server to the Python package which demonstrates publishing Protobuf data.
- Adds a new installation option which installs the protobuf dependency — this is only required if running the example server, not for other uses of the foxglove-websocket package. Install via: `pip install foxglove-websocket[examples]`

Closes https://github.com/foxglove/ws-protocol/issues/27


python/src/foxglove_websocket/examples/protobuf_server.py
```diff
@@ -0,0 +1,73 @@
...
+
+- `ExampleMsg.proto`: hand-written Protobuf schema that describes data the server will send.
+
+- `ExampleMsg_pb2.py`: generated Python code from `protoc --python_out=. ExampleMsg.proto`. The
+  example server uses this to serialize messages as binary data via SerializeToString().
+
+- `ExampleMsg.bin`: generated FileDescriptorSet from `protoc --descriptor_set_out=ExampleMsg.bin
+  ExampleMsg.proto`. This binary blob gets passed to clients as the `schema`; Foxglove Studio uses
+  this along with the `schemaName` to decode the message data. A FileDescriptorSet can represent
+  multiple input .proto files (see the --include_imports option to protoc).
```
jtbandes, line 14: @esthersweon Not sure if this is better placed in the README? The point of this example is for people to actually use it as a starting point for their own servers, so I expect people will come here and see this docstring...

python/README.md
```diff
@@ -10,70 +10,25 @@ The protocol is encoding-agnostic, i.e. it can support Protobuf messages, JSON m
...
+```
 
 The following script provides a simple example server which publishes messages on a single topic called `example_msg`, using JSON to encode message data and [JSON Schema](https://json-schema.org) to describe the message layout.
 
 To get started with the example server:
 
-1. Copy the script below into a file named `example.py` and run `python3 example.py`.  
-   **or:** Install the foxglove-websocket package and run `python -m foxglove_websocket.examples.simple_server`.
+1. Run `python -m foxglove_websocket.examples.json_server`.  
+   **or:** Run `python -m foxglove_websocket.examples.protobuf_server`.
```
jtbandes, line 26: I expect this will change in #24.

amacneil:
awesome

PR #39, opened 2022-01-12
cpp: Add removeChannel, missing include, operator==

jtbandes:
A couple of API and build improvements.

### DIFF
--- a/cpp/foxglove-websocket/include/foxglove/websocket/server.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/server.hpp
@@ -42,7 +42,7 @@ struct ChannelWithoutId {
 struct Channel : ChannelWithoutId {
   ChannelId id;
 
-  explicit Channel(ChannelId id, ChannelWithoutId&& ch)
+  explicit Channel(ChannelId id, ChannelWithoutId ch)
       : ChannelWithoutId(std::move(ch))
       , id(id) {}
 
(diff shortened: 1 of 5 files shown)


cpp/foxglove-websocket/include/foxglove/websocket/server.hpp
```diff
@@ -31,11 +32,17 @@ struct ChannelWithoutId {
   std::string encoding;
   std::string schemaName;
   std::string schema;
+
+  bool operator==(const ChannelWithoutId& other) const {
+    return topic == other.topic && encoding == other.encoding && schemaName == other.schemaName &&
```
KerstinKeller, line 37: just do a 
```
bool operator==(ChannelWithoutId const& rhs) const = default;
```
It should be C++11 and the compiler generates just that.

cpp/foxglove-websocket/include/foxglove/websocket/server.hpp
```diff
@@ -48,6 +55,10 @@ struct Channel : ChannelWithoutId {
       {"schema", channel.schema},
     };
   }
+
+  bool operator==(const Channel& other) const {
+    return id == other.id && ChannelWithoutId::operator==(other);
```
KerstinKeller, line 60: Same as above?

cpp/foxglove-websocket/include/foxglove/websocket/server.hpp
jtbandes: Hmm, seems this is a c++20 feature? never heard of it before :) https://en.cppreference.com/w/cpp/language/default_comparisons

KerstinKeller: I could have sworn that I used it before, but apparently not... too bad 😞

Issue #38, opened 2022-01-07
send message from client to server


0wu:
At the moment data flow is purely from server to client.
However, in actual robotics environment, user may want to send control signals from the UI to server.
Such feature (eg. Publish) is supported only when using native ROS.
I would like to see this happening on ws-protocol as well.

jtbandes:
Thanks for your interest! I agree this would be valuable. The [rosbridge](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md#341-advertise--advertise-) server has this feature already so we can look there for inspiration.

Here are some notes and thoughts on how we might implement it. Let me know if you have any feedback.

- Add `"advertise"` (or `"clientAdvertise"`? name TBD) to server capabilites
- Allow clients to send Advertise / Unadvertise / Message Data ops to server
  - Would Message Data still have to be binary, or do we allow clients to send JSON?
- Should the server explicitly Subscribe / Unsubscribe from client messages, or should the client send them whenever it wants to?
  - The use cases I'm familiar with are low-frequency messages sent from the client, such as sending a message when a user clicks a button. So there's not much benefit to reducing traffic by only sending messages if the server explicitly subscribed. (The goal of the Foxglove WebSocket protocol is not to be a full-featured symmetrical pub/sub middleware.) We would need to understand some motivating use cases in order to justify requiring the server to subscribe/unsubscribe.

0wu:
Our use-cases
* some keyboard based teleoperation controlling a robot arm or vehicle, 10Hz 
* start/stop experiment, one-off
So json is definitely suitable.

Our server side system does not rely on ROS for IPC but we use ROS message format for SerDes. Foxglove plays the visualization/ui role here. That's why we can't use native ROS interface.  Maybe implementing our own `rosbridge-like` is one way to go. However, this web-socket protocol seems rather light weight.

sionyx:
Looking forward for this feature as well. My system not based on ROS, but I wrote rosbridge protocol based bridge to control the system. At this time I have the bridge migrated to foxglove ws-protocol and realized that I cannot send any data back. 

It doesn't matter for my bridge to have or not Subscribe / Unsubscribe operations, it can receive messages from client at any time. Json format is ok as well.

sionyx:
Hey guys! Are there any news regarding this feature?

jtbandes:
[...]

PR #158, opened 2022-05-23
Added publishing ability to FoxgloveClient

saryazdi:
* Added publishing ability to FoxgloveClient: `advertise()`, `unadvertise()`, `sendData()`
* Updated python server examples to handle messages from client

**Public-Facing Changes**
Ability to publish from client (Foxglove studio) to server (e.g. robot) through the Publish/Teleop/etc panels.

**Demo Steps**
See attached video for final result:

https://user-images.githubusercontent.com/33838550/170392225-3d9fc846-f814-4a31-b45d-e7407e2c2293.mp4

1. Pull corresponding branch from [foxglove/studio PR](https://github.com/foxglove/studio/pull/3431) 
2. Launch foxglove client: `yarn install && yarn:serve`
3. Pull this branch (foxglove/ws-protocol) and launch any of the python servers, e.g.: `python python/src/foxglove_websocket/examples/json_server.py` 
4. In foxglove client, publish data through Teleop/Publish panels.
5. The active python server console should print the messages being received.

**Description**
* Should be merged before https://github.com/foxglove/studio/pull/3431
* Addresses https://github.com/foxglove/ws-protocol/issues/38

**Future TODOs**
- update c++/typescript server examples.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -234,7 +234,9 @@ Informs the server that client channels are no longer available.
 {
   "op": "clientData",
   "topic": "foo",
-  "data": { "ExampleData": [1, 2] },
+  "data": {
+    "ExampleData": [1, 2]
+  },
   "timestamp": 123456789
 }
 ```
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -33,6 +33,9 @@
 
 - [Subscribe](#subscribe) (json)
 - [Unsubscribe](#unsubscribe) (json)
+- [Client Advertise](#client-advertise) (json)
+- [Client Unadvertise](#client-unadvertise) (json)
+- [Client Data](#client-data) (json)
 
 ## JSON messages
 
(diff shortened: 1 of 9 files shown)


typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -110,6 +112,53 @@ export default class FoxgloveClient {
...
+   * Remove a previously advertised channel and inform any connected servers.
+   */
+  removeChannel(channelId: ChannelId): void {
+    if (!this.channelsById.delete(channelId)) {
+      throw new Error(`Channel ${channelId} does not exist`);
+    }
+    this.send({ op: "unadvertise", channelIds: [channelId] });
+  }
+
+  sendMessageData(subscriptionId: SubscriptionId, timestamp: bigint, payload: BufferSource): void {
```
jtbandes, line 137: From the discussion on #38, it sounds like JSON would be sufficient for a first pass implementation of publishing. So I would change this to accept an object instead, something like this:
```suggestion
  sendMessage(subscriptionId: SubscriptionId, timestamp: bigint, payload: Record<string, unknown>): void {
```
And we can send a JSON message rather than a binary message on the websocket.

jtbandes, line 137: From the discussion on #38, it sounds like JSON would be sufficient for a first pass implementation of publishing. So I would change this to accept an object instead, something like this:
```suggestion
  sendMessage(subscriptionId: SubscriptionId, timestamp: bigint, payload: Record<string, unknown>): void {
```
And we can send a JSON message rather than a binary message on the websocket.

saryazdi:
@jtbandes @jhurliman gentle reminder. thank you!

docs/spec.md
```diff
@@ -170,6 +173,74 @@ Informs the client that channels are no longer available.
...
 
+### Client Advertise
+
+- Informs the server about available client channels.
+
+#### Fields
+
+- `op`: string `"clientAdvertise"`
+- `channels`: array of:
+  - `topic`: string
```
jtbandes, line 184: I think we should include a channel id, so that the full topic name doesn't have to be provided each time a message is published

jtbandes (commented):
Thanks for your patience and for updating the docs as well! Here are some design suggestions. Let me know whether you would like to work on these, or if you don't have bandwidth and need to hand off the project to us for completion.

python/src/foxglove_websocket/server/__init__.py
```diff
@@ -210,7 +254,7 @@ async def _handle_connection(
                 {
                     "op": "serverInfo",
                     "name": self.name,
-                    "capabilities": [],
+                    "capabilities": [ServerCapabilities.receiveClientData.value],
```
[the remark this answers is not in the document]
saryazdi: I was thinking a live foxglove websocket server would always support receiving data (I think that would make it similar to Rosbridge?). I can also change it, E.g.:
```python
class FoxgloveServer:
    ...
    def __init__(
            self,
            host: str,
            port: Optional[int],
            name: str,
            *,
            receives: bool = True,
            logger: logging.Logger = _get_default_logger(),
        ):
```
lmk what you think.

docs/spec.md
```diff
@@ -170,6 +173,74 @@ Informs the client that channels are no longer available.
...
+```
+
+### Client Data
+
+- A message sent from client to the server.
+- Client messages are in json encoding.
+
+#### Fields
+
+- `op`: string `"clientData"`
```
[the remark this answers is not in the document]
saryazdi: Originally went w/ `clientData` to be consistent w/ `ServerCapabilities.receiveClientData`. Changed it to `clientMessageData` now. Let me know if you want me to switch back (easy fix).

jhurliman:
Now that #250 is merged, I'm working on PRs for one language at a time to support the new "clientPublish" capability. TypeScript PR is here: https://github.com/foxglove/ws-protocol/pull/276

jhurliman:
Closing in favor of #276. The Python and C++ code will need to be updated in followup PRs.

PR #250, opened 2022-10-21
Extend spec with ability to send data from client to server.

### LINKED ISSUE #38: send message from client to server

achim-k:
**Public-Facing Changes**
Extend the spec with ability to send data from client to server.

**Description**
This PR adds 3 new operations to the spec to allow sending data from the client to the server. Once we have agreed on the spec, I will update the example server implementation accordingly.

Partially fixes #38 
Partially supersedes #158

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -35,7 +35,7 @@
 - [Unsubscribe](#unsubscribe) (json)
 - [Client Advertise](#client-advertise) (json)
 - [Client Unadvertise](#client-unadvertise) (json)
-- [Client Message Data](#client-message-data) (json)
+- [Client Message Data](#client-message-data) (binary)
 
 ## JSON messages
 
@@ -223,32 +223,15 @@ Informs the client that channels are no longer available.
 
 ### Client Publish
 
-- Sends a JSON message from the client to the server. Note that the client is only allowed to publish messages if the server previously declared that it has the `clientPublish` [capability](#server-info).
+- Sends a binary websocket message that contains the JSON encoded messsage from the client to the server. Note that the client is only allowed to publish messages if the server previously declared that it has the `clientPublish` [capability](#server-info).
 
 #### Fields
 
-- `op`: string `"publish"`
-- `channelId`: number. Channel ID corresponding to previous [Client Advertise](#client-advertise)
-- `data`: object. JSON object
-
-#### Example
-
-```json
-{
-  "op": "publish",
-  "channelId": 1,
-  "data": {
-    "header": {
-      "frame_id": "/map"
-    },
-    "point": {
-      "x": 1.0,
-      "y": 2.0,
-      "z": 0.0
-    }
-  }
-}
-```
+| Bytes           | Type    | Description                     |
+| --------------- | ------- | ------------------------------- |
+| 1               | opcode  | 0x01                            |
+| 4               | uint32  | channel id                      |
+| remaining bytes | uint8[] | message payload (JSON enconded) |
 
 ## Binary messages
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -184,7 +184,7 @@ Informs the client that channels are no longer available.
 - `channels`: array of:
   - `id`: number chosen by the client. The client may reuse ids that have previously been unadvertised.
   - `topic`: string
-  - `encoding`: string, must be `"json"`
+  - `encoding`: string
   - `schemaName`: string
 
 #### Example
@@ -196,7 +196,7 @@ Informs the client that channels are no longer available.
     {
       "id": 1,
       "topic": "foo",
-      "encoding": "json",
+      "encoding": "protobuf",
       "schemaName": "ExampleMsg"
     }
   ]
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -170,6 +173,82 @@ Informs the client that channels are no longer available.
...
+
+- Informs the server about available client channels.
+
+#### Fields
+
+- `op`: string `"client_advertise"`
+- `channels`: array of:
+  - `id`: number chosen by the client. The client may not reuse ids across multiple advertised channels.
+  - `topic`: string
+  - `encoding`: string `"json"`
```
jtbandes, line 186:
```suggestion
  - `encoding`: string, must be `"json"`
```

docs/spec.md
```diff
@@ -170,6 +173,82 @@ Informs the client that channels are no longer available.
...
+#### Example
+
+```json
+{
+  "op": "client_advertise",
+  "channels": [
+    {
+      "id": 1,
+      "topic": "foo",
+      "encoding": "json",
```
jtbandes, line 198: If we wanted to expand to binary publishing in the future, is this sufficient? Would we need to distinguish between a json websocket message and a binary websocket message that contains json-encoded data?

achim-k: I think in the future we can also allow the client to send message data in binary form. This could e.g. look similar to the binary messages send by the server: 

| Bytes           | Type    | Description                     |
| --------------- | ------- | ------------------------------- |
| 1               | opcode  | 0x01                            |
| 4               | uint32  | channel id                     |
| remaining bytes | uint8[] | message payload                 |

The first byte of the incoming message would tell the server if it's an operation message (JSON encoded) or if it is a binary message. If it is a binary message, the channel `encoding` would tell the server how to interpret the message payload. So if it's `json`, the server would parse the payload as JSON.

docs/spec.md
```diff
@@ -170,6 +173,82 @@ Informs the client that channels are no longer available.
 }
 ```
 
+### Client Advertise
+
+- Informs the server about available client channels.
+
+#### Fields
+
+- `op`: string `"client_advertise"`
```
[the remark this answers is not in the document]
achim-k: > On the other hand, what do you think about using `"op": "advertise"`? Since the message is sent from client to server, there's no ambiguity. Similarly ops `unadvertise` and `message_data` (or just `message`?) below.

I like your proposal. Maybe we could name the sending message operation `publish` (similar to rosbridge). This would be more consistent with all operations being verbs.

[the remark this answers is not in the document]
jtbandes: > all operations being verbs

not really true for the server->client message op, but since it's a binary op, the opcode is just `1` :)

`publish` is fine.

docs/spec.md
```diff
@@ -170,6 +174,82 @@ Informs the client that channels are no longer available.
 }
 ```
 
+### Client Advertise
+
+- Informs the server about available client channels. Note that hhe client is only allowed to advertise channels if the server previously declared that it has the `clientPublish` [capability](#server-info).
```
jtbandes, line 179:
```suggestion
- Informs the server about available client channels. Note that the client is only allowed to advertise channels if the server previously declared that it has the `clientPublish` [capability](#server-info).
```

jtbandes: I'd also add this note to the other 2 new ops

docs/spec.md
```diff
@@ -170,6 +174,82 @@ Informs the client that channels are no longer available.
...
+
+#### Fields
+
+- `op`: string `"advertise"`
+- `channels`: array of:
+  - `id`: number chosen by the client. The client may not reuse ids across multiple advertised channels.
```
jtbandes, line 185: We said this about channelId for server advertise:

> The server may reuse ids when channels disappear and reappear, but only if the channel keeps the exact same topic, encoding, schemaName, and schema. Clients will use this unique id to cache schema info and deserialization routines.

Should we do something like this here too? Currently it doesn't sound like the same id is allowed to be reused even if it's unadvertised in between? Keeping the restriction as you have it is fine too, but might want to make it more clear that even unadvertising does not allow you to use the same id again.

jtbandes (approved):

docs/spec.md
achim-k: We can allow the client to reuse channel IDs if they have been unadvertised in between. I see no problem in that as the server will use the channel ID only to lookup published channels of the client

docs/spec.md
```diff
@@ -170,6 +174,82 @@ Informs the client that channels are no longer available.
...
+#### Example
+
+```json
+{
+  "op": "unadvertise",
+  "channelIds": [1, 2]
+}
+```
+
+### Client Message Data
```
jtbandes, line 224: Nit: maybe rename this to "client publish" to match the `op`? Or rename the `op` to `"msg"` or similar?

jtbandes (approved):

docs/spec.md
jhurliman, line 224:
```suggestion
### Client Publish
```

docs/spec.md
```diff
@@ -170,6 +174,82 @@ Informs the client that channels are no longer available.
...
+
+- Sends a JSON message from the client to the server. Note that the client is only allowed to publish messages if the server previously declared that it has the `clientPublish` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"publish"`
+- `channelId`: number. Channel ID corresponding to previous [Client Advertise](#client-advertise)
+- `data`: object. JSON object
```
jhurliman, line 232: I think this should be a binary message for protocol simplicity. For this one example encoding of JSON a non-binary message works, but most encodings will require binary. It simplifies the server code if we don't need to handle client-published messages with two separate code paths.

jhurliman (changes requested):
I think we should change client-published messages to a binary WebSocket message. The rest LGTM

docs/spec.md
jtbandes: > For this one example encoding of JSON

As currently written in the PR, JSON is the only allowed encoding. The thinking was that it would add a lot of complexity to ask clients to learn how to encode messages in multiple encodings, and somehow discover which encodings the server would support. How would you recommend we address these issues?

jhurliman: > add a lot of complexity to ask clients to learn how to encode messages in multiple encodings

I'm not suggesting clients should support multiple encodings. If a ws-protocol client is doing bag playback of a ROS 1 bag, it makes sense to advertise ros1-encoded topics and publish unprocessed message payloads rather than requiring the client to decode, transform to JSON, then publish. In this example, the client would only support publishing ros1-encoding and if the server did not support that encoding it would not be able to use the publish capability.

This does require the server to advertise which encodings it supports for client publishing. I'm not sure where that slots into the protocol best.

jhurliman (approved):

Issue #255, opened 2022-10-25
Update examples to dynamically generate FileDescriptorSet


jtbandes:
Nowadays, rather than suggesting `protoc --descriptor_set_out`, we point people at a technique to generate the FileDescriptorSet at runtime, which is used in mcap libraries/examples:

https://github.com/foxglove/mcap/blob/21b737f99800bfe1d733b769022a726ab3ca4e0e/python/mcap-protobuf-support/mcap_protobuf/schema.py#L17

https://github.com/foxglove/mcap/blob/21b737f99800bfe1d733b769022a726ab3ca4e0e/cpp/examples/protobuf/writer.cpp#L48

We should update the examples in this repo to use the technique.

PR #269, opened 2022-11-09
Update python and c++ protobuf examples to demonstrate dynamic FileDescriptorSet generation

### LINKED ISSUE #138: Add C++ Protobuf example
We should provide example code for using Protobuf encoding in C++.

This will also require importing or adding our own base64 encoding implementation.

jtbandes:
**Public-Facing Changes**
Updated example Python Protobuf server to demonstrate dynamic generation of FileDescriptorSet schemas.
Updated example C++ server to use Protobuf instead of JSON, including dynamic generation of FileDescriptorSet schemas.

**Description**
Resolves https://github.com/foxglove/ws-protocol/issues/255
Resolves #138
Relates to https://github.com/foxglove/mcap/pull/694

Checks in generated files for Python because we have not yet published them in a Python package (https://github.com/foxglove/schemas/issues/71).

https://user-images.githubusercontent.com/14237/200960945-1b08047f-38d9-49c3-85d5-bd84d7fdec67.mov


defunctzombie (approved):
What's a few more hundred files between friends?

Committing the files does seem weird from the perspective of every other protobuf codebase I've seen out there. So if they are hard to generate maybe thats by design since you are meant to generate them with whatever build system you've selected?

jtbandes:
It's not that they are hard to generate, but they are hard to generate _automatically_ with CMake (and I didn't even explore generation as part of the python build process). With python I think a nice path forward would be https://github.com/foxglove/schemas/issues/71 but didn't want to block this PR on that. For C++ I am not sure what other options we have...I suppose we could similarly publish a conan package with these files.

defunctzombie:
> For C++ I am not sure what other options we have...I suppose we could similarly publish a conan package with these files.

Isn't it common to generate these in your project? Why would we publish the generated files to conan?

jtbandes:
Hm, I see we already have some of the CMake magic figured out in https://github.com/foxglove/mcap/blob/main/cpp/examples/protobuf/CMakeLists.txt. I will have to see if that can be translated.

PR #270, opened 2022-11-11
Add mcap-play example

jtbandes:
**Public-Facing Changes**
Added mcap-play example which plays a local MCAP file over the websocket.
```
$ npx @foxglove/ws-protocol-examples@latest mcap-play <file> [--rate 1.0] [--loop]
```


typescript/ws-protocol-examples/yarn-error.log
```diff
@@ -0,0 +1,4146 @@
+Arguments:
```
jhurliman, line 1: Accidental?

jhurliman (approved):
I think this example clearly demonstrates how much boilerplate is required to use the MCAP TypeScript library

PR #276, opened 2022-11-16
Implement client publishing in the TypeScript client and server

jhurliman:
**Public-Facing Changes**
- TypeScript client and server support for client-advertised topics and client-published messages

**Description**

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveServer.ts
+++ b/typescript/ws-protocol/src/FoxgloveServer.ts
@@ -5,7 +5,9 @@ import { ChannelId, StatusLevel } from ".";
 import {
   BinaryOpcode,
   Channel,
+  ClientBinaryOpcode,
   ClientChannel,
+  ClientChannelId,
   ClientMessage,
   ClientPublish,
   IWebSocket,
@@ -19,7 +21,7 @@ type ClientInfo = {
   connection: IWebSocket;
   subscriptions: Map<SubscriptionId, ChannelId>;
   subscriptionsByChannel: Map<ChannelId, Set<SubscriptionId>>;
-  advertisements: Map<ChannelId, ClientChannel>;
+  advertisements: Map<ClientChannelId, ClientChannel>;
 };
 
 type SingleClient = { client: ClientInfo };
@@ -169,7 +171,17 @@ export default class FoxgloveServer {
     connection.onmessage = (event: MessageEvent<ArrayBuffer | string>) => {
       if (typeof event.data === "string") {
         // TEXT (JSON) message handling
-        const message = JSON.parse(event.data) as unknown;
+        let message: unknown;
+        try {
+          message = JSON.parse(event.data) as unknown;
+        } catch (error) {
+          this.emitter.emit(
+            "error",
+            new Error(`Invalid JSON message from ${name}: ${(error as Error).message}`),
+          );
+          return;
+        }
+
         if (typeof message !== "object" || message == undefined) {
           this.emitter.emit("error", new Error(`Expected JSON object, got ${typeof message}`));
           return;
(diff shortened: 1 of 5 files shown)


typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -236,11 +264,47 @@ export default class FoxgloveServer {
 
         break;
 
+      case "advertise":
+        for (const channel of message.channels) {
+          client.advertisements.set(channel.id, channel);
```
jtbandes, line 269: do we want to check if there are any existing channels for this id and make sure the client isn't overwriting them?

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -3,25 +3,33 @@ import { EventEmitter, EventNames, EventListener } from "eventemitter3";
...
 
 type EventTypes = {
+  error: (error: Error) => void;
+
   subscribe: (channel: ChannelId) => void;
   unsubscribe: (channel: ChannelId) => void;
-  error: (error: Error) => void;
+  message: (event: ClientPublish) => void;
+  advertise: (channel: ClientChannel) => void;
+  unadvertise: (channel: ChannelId) => void;
```
jtbandes, line 24: I wonder if naming these `clientAdvertise`/`clientUnadvertise` would help reduce confusion? On the other hand, we didn't name the other ones `clientSubscribe`/`clientUnsubscribe`....

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -3,25 +3,33 @@ import { EventEmitter, EventNames, EventListener } from "eventemitter3";
...
 };
 
 type ClientInfo = {
   name: string;
   connection: IWebSocket;
   subscriptions: Map<SubscriptionId, ChannelId>;
   subscriptionsByChannel: Map<ChannelId, Set<SubscriptionId>>;
+  advertisements: Map<ChannelId, ClientChannel>;
```
jtbandes, line 32: nit: maybe make a separate type for ClientChannelId since it is a distinct namespace from server channel ids? (Ideally we should use tagged/branded types to actually prevent misuse...)

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -145,17 +160,30 @@ export default class FoxgloveServer {
...
-      if (typeof message !== "object" || message == undefined) {
-        throw new Error(`Expected JSON object, got ${typeof message}`);
-      }
-      try {
-        this.handleClientMessage(client, message as ClientMessage);
-      } catch (error) {
-        this.emitter.emit("error", error as Error);
+      if (typeof event.data === "string") {
+        // TEXT (JSON) message handling
+        const message = JSON.parse(event.data) as unknown;
```
jtbandes, line 165: Not new in your PR, but it occurs to me that maybe the `JSON.parse` should be inside a try/catch as well. We might end up with an uncaught exception if the client sends invalid json.

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -236,11 +264,47 @@ export default class FoxgloveServer {
...
+
       default:
         throw new Error(`Unrecognized client opcode: ${(message as { op: string }).op}`);
     }
   }
 
+  private handleClientBinaryMessage(client: ClientInfo, message: DataView): void {
+    const opcode = message.getUint8(0);
+    switch (opcode) {
+      case BinaryOpcode.MESSAGE_DATA: {
```
jtbandes, line 291: nit: consider separating client binary opcodes from server binary opcodes? Since the format of the message is not identical.

jtbandes (approved):
🎉

PR #277, opened 2022-11-20
python readme: string literal for [examples]

alkasm:
zsh (the default shell on MacOS) needs backslashes to be escaped or the argument to be quoted, see e.g.: https://stackoverflow.com/a/30539963/5087436


jtbandes (approved):
Thanks!

PR #280, opened 2022-11-23
Add minimum nodejs version requirement

jtbandes:
**Public-Facing Changes**
Added minimum NodeJS version requirement for `npx @foxglove/ws-protocol-examples`

**Description**
A user reported an error using an old version of Node.


defunctzombie (approved):
👍 soon 14 will be EOL

PR #288, opened 2022-12-02
Extend spec with parameter support

achim-k:
**Public-Facing Changes**
Extend spec with parameter support

**Description**
Adds the following new capabilities to the spec

  - `parameters`: Allow clients to get & set parameters
  - `parameterSubscribe`: Allow clients to subscribe to parameter changes

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -341,11 +341,11 @@ Set one or more parameters. Only supported if the server previously declared tha
 
 Subscribe to parameter updates. Only supported if the server previously declared that it has the `parameterSubscribe` [capability](#server-info).
 
-Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
+Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
 
 #### Fields
 
-- `op`: string `"subscribeParameterUdpates"`
+- `op`: string `"subscribeParameterUpdates"`
 - `parameterNames`: string[], leave empty to subscribe to all currently known parameters
 
 #### Example
@@ -368,7 +368,7 @@ Unsubscribe from parameter updates. Only supported if the server previously decl
 
 #### Fields
 
-- `op`: string `"unsubscribeParameterUdpates"`
+- `op`: string `"unsubscribeParameterUpdates"`
 - `parameterNames`: string[], leave empty to unsubscribe from all parameter updates
 
 #### Example
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -323,7 +323,7 @@ Set one or more parameters. Only supported if the server previously declared tha
 
 Subscribe to parameter updates. Only supported if the server previously declared that it has the `parametersSubscribe` [capability](#server-info).
 
-Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
+Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUpdates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
 
 #### Fields
 
@@ -334,7 +334,7 @@ Sending `subscribeParameterUpdates` multiple times will append the list of param
 
 ```json
 {
-  "op": "subscribeParameterUdpates",
+  "op": "subscribeParameterUpdates",
   "parameterNames": [
     "/int_param",
     "/float_param",
@@ -357,7 +357,7 @@ Unsubscribe from parameter updates. Only supported if the server previously decl
 
 ```json
 {
-  "op": "unsubscribeParameterUdpates",
+  "op": "unsubscribeParameterUpdates",
   "parameterNames": [
     "/int_param",
     "/float_param",
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -143,18 +143,17 @@ Informs the client that channels are no longer available.
 
 Informs the client about parameters. Only supported if the server declares the `parameters` [capability](#server-info).
 
-The server may send this message
-
-- at arbitrary points in time
-- as a response to the client's [Get Parameters request](#get-parameters)
-- when parameters have changed which were [subscribed](#subscribe-parameter-update) by the client
-
 #### Fields
 
 - `op`: string `"parameterValues"`
 - `parameters`: array of:
   - `name`: string, name of the parameter
   - `value`: number | boolean | string | number[] | boolean[] | string[]
+- `type`: string, one of
+  - `"serverInitiated"` - The message was sent by the server without being previously requested
+  - `"getParametersResponse"` - The message was sent as a response to the client's [Get Parameters request](#get-parameters)
+  - `"valueUpdate"` - The message was sent because one or more [subscribed](#subscribe-parameter-update) parameters have changed their value
+- `id`: string | undefined. Only set when `type` is `"getParametersResponse"` and the [request's](#get-parameters) `id` field was set
 
 #### Example
 
@@ -166,7 +165,9 @@ The server may send this message
     { "name": "/float_param", "value": 1.2 },
     { "name": "/string_param", "value": "foo" },
     { "name": "/node/nested_ints_param", "value": [1, 2, 3] }
-  ]
+  ],
+  "type": "getParametersResponse",
+  "id": "request-123"
 }
 ```
 
@@ -281,6 +282,7 @@ Request one or more parameters. Only supported if the server previously declared
 
 - `op`: string `"getParameters"`
 - `parameterNames`: string[], leave empty to retrieve all currently set parameters
+- `id`: string | undefined, arbitrary string used for identifying the corresponding server [response](#parameter-values)
 
 #### Example
 
@@ -292,7 +294,8 @@ Request one or more parameters. Only supported if the server previously declared
     "/float_param",
     "/string_param",
     "/node/nested_ints_param"
-  ]
+  ],
+  "id": "request-123"
 }
 ```
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -28,6 +28,8 @@
 - [Advertise](#advertise) (json)
 - [Unadvertise](#unadvertise) (json)
 - [Message Data](#message-data) (binary)
+- [Parameter Values](#parameter-values) (json)
+- [Parameter Updates](#parameter-updates) (json)
 
 ### Sent by client
 
@@ -36,6 +38,10 @@
 - [Client Advertise](#client-advertise) (json)
 - [Client Unadvertise](#client-unadvertise) (json)
 - [Client Message Data](#client-message-data) (binary)
+- [Get Parameters](#get-parameters) (json)
+- [Set Parameters](#set-parameters) (json)
+- [Subscribe Parameter Update](#subscribe-parameter-update) (json)
+- [Unsubscribe Parameter Update](#unsubscribe-parameter-update) (json)
 
 ## JSON messages
 
@@ -51,6 +57,7 @@ Each JSON message must be an object containing a field called `op` which identif
 - `name`: free-form information about the server which the client may optionally display or use for debugging purposes
 - `capabilities`: array of strings, informing the client about which optional features are supported
   - `clientPublish`: Allow clients to advertise channels to send data messages to the server
+  - `parameters`: Allow clients to get & set parameters
 
 #### Example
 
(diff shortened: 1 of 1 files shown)


achim-k:
@jtbandes @jhurliman PTAL at this early draft, thanks!

docs/spec.md
```diff
@@ -132,6 +139,66 @@ Informs the client that channels are no longer available.
...
+
+Informs the client about updates of parameters that the client previously [subscribed](#subscribe-parameter-update) to. Only supported if the server declares the `parameters` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"parameterUpdates"`
+- `parameters`: array of:
+  - `name`: string
+  - `newValue`: number | boolean | string | number[] | boolean[] | string[]
+  - `oldValue`: number | boolean | string | number[] | boolean[] | string[]
```
jtbandes, line 183: Why include oldValue?

docs/spec.md
```diff
@@ -132,6 +139,66 @@ Informs the client that channels are no longer available.
...
+This message is sent as a response to the client's [Get Parameters request](#get-parameters). Only supported if the server declares the `parameters` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"parameterValues"`
+- `parameters`: array of:
+  - `name`: string, name of the parameter
+  - `value`: number | boolean | string | number[] | boolean[] | string[]
+  - `meta`: object, only given if requested with `includeMetaInfo: true`
+    - `type`: bool | int64 | float64 | string | byte[] | bool[] | int64[] | float64[] | string[]
```
jtbandes, line 153: Do you think all systems that might want to implement parameter support will have this same kind of meta information? Why bother making it optional? Is the type info redundant with the `value` since you can just check the JSON type? (I guess the specific type may matter for numbers, but Im' not sure what the use case is)

nit: the possible values for `type` should all be wrapped in quotes, since they are strings, right? or just "`type`: `string`"?

Why bother nesting type under `meta`? Are there other things we will want to include in `meta`?

docs/spec.md
```diff
@@ -132,6 +139,66 @@ Informs the client that channels are no longer available.
 }
 ```
 
+### Parameter Values
+
+This message is sent as a response to the client's [Get Parameters request](#get-parameters). Only supported if the server declares the `parameters` [capability](#server-info).
```
jtbandes, line 144: Can the server choose to send parameters on its own or does the client have to request them?

docs/spec.md
```diff
@@ -132,6 +139,66 @@ Informs the client that channels are no longer available.
...
+      "value": [1, 2, 3],
+      "meta": { "type": "int64[]" }
+    }
+  ]
+}
+```
+
+### Parameter Updates
+
+Informs the client about updates of parameters that the client previously [subscribed](#subscribe-parameter-update) to. Only supported if the server declares the `parameters` [capability](#server-info).
```
jtbandes, line 175: Do you think all systems that might want to implement parameter support will have support for subscribing to updates?

docs/spec.md
```diff
@@ -235,6 +302,101 @@ Informs the client that channels are no longer available.
...
+```
+
+### Subscribe Parameter Update
+
+Subscribe to parameter updates. Only supported if the server previously declared that it has the `parameters` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"subscribeParameterUdpates"`
+- `parameterNames`: string[], leave empty to subscribe to all currently known parameters
```
jtbandes, line 361: Is this meant to replace the existing set of subscriptions or append/merge?

docs/spec.md
achim-k: With the rosbridge protocol, the client has to actively request paramters (via a service call), that's why I opted for that mechanism. But you are right, I see no reason why the server can not also publish available parameters without being requested to do so. If we allow this, I will delete the first sentence of this paragraph

docs/spec.md
achim-k: > Do you think all systems that might want to implement parameter support will have this same kind of meta information?

The meta field is probably very system specific. I intended it as a struct where the server could communicate additional parameter information such as the type, description, min / max / default values, etc. I'm not sure if we should enforce a schema on that or if we keep it schema-less. In the end, the `meta` info is optional, clients will still be able to display parameter values without interpreting it

docs/spec.md
achim-k: Good remark. No, I don't think so. I guess the server will have to declare a separate `parameter-update` capability for that.

docs/spec.md
achim-k: Not necessary, I agree

```suggestion
```

docs/spec.md
achim-k: It is meant as append/merge. You will be able to subscribe to a single parameter at max once. Will add a comment about that

docs/spec.md
jhurliman: Yes, I think a separate capability should be added (I think we're using camel case for capabilities?). Parameter update sounds like the capability to write to parameters, can we call it `parameterSubscribe`?

docs/spec.md
jhurliman: Drop `oldValue`, rename `newValue` to `value`

docs/spec.md
```diff
@@ -235,6 +302,101 @@ Informs the client that channels are no longer available.
...
 
+### Get Parameters
+
+Request one or more parameters. Only supported if the server previously declared that it has the `parameters` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"getParameters"`
+- `parameterNames`: string[], leave empty to retrieve all currently set parameters
+- `includeMetaInfo`: boolean, set to true to retrieve additional meta information for each parameter
```
jhurliman, line 313: Remove `includeMetaInfo`

docs/spec.md
```diff
@@ -235,6 +302,101 @@ Informs the client that channels are no longer available.
...
+  "parameters": [
+    { "name": "/int_param", "value": 3 },
+    { "name": "/float_param", "value": 4.1 }
+  ]
+}
+```
+
+### Subscribe Parameter Update
+
+Subscribe to parameter updates. Only supported if the server previously declared that it has the `parameters` [capability](#server-info).
```
jhurliman, line 356: `parameterSubscribe` capability

docs/spec.md
```diff
@@ -235,6 +302,101 @@ Informs the client that channels are no longer available.
...
+    "/float_param",
+    "/string_param",
+    "/node/nested_ints_param"
+  ]
+}
+```
+
+### Unsubscribe Parameter Update
+
+Unsubscribe from parameter updates. Only supported if the server previously declared that it has the `parameters` [capability](#server-info).
```
jhurliman, line 379: `parameterSubscribe`

jhurliman (changes requested):

docs/spec.md
```diff
@@ -51,6 +57,8 @@ Each JSON message must be an object containing a field called `op` which identif
 - `name`: free-form information about the server which the client may optionally display or use for debugging purposes
 - `capabilities`: array of strings, informing the client about which optional features are supported
   - `clientPublish`: Allow clients to advertise channels to send data messages to the server
+  - `parameters`: Allow clients to get & set parameters
+  - `parameterSubscribe`: Allow clients to subscribe to parameter changes
```
achim-k, line 61: Should we call it just `parameter` to be inline with `parameterSubscribe` ?

```suggestion
  - `parameter`: Allow clients to get & set parameters
  - `parameterSubscribe`: Allow clients to subscribe to parameter changes
```

jhurliman: Or maybe `parameterSubscribe` should be `parametersSubscribe`. You make the call, change one of them so they're consistent.

docs/spec.md
```diff
@@ -235,6 +291,100 @@ Informs the client that channels are no longer available.
...
+    { "name": "/float_param", "value": 4.1 }
+  ]
+}
+```
+
+### Subscribe Parameter Update
+
+Subscribe to parameter updates. Only supported if the server previously declared that it has the `parameterSubscribe` [capability](#server-info).
+
+Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
```
jhurliman, line 344:
```suggestion
Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
```

docs/spec.md
```diff
@@ -235,6 +291,100 @@ Informs the client that channels are no longer available.
...
+
+#### Fields
+
+- `op`: string `"subscribeParameterUdpates"`
```
jhurliman, line 348:
```suggestion
- `op`: string `"subscribeParameterUpdates"`
```

docs/spec.md
```diff
@@ -235,6 +291,100 @@ Informs the client that channels are no longer available.
...
+}
+```
+
+### Unsubscribe Parameter Update
+
+Unsubscribe from parameter updates. Only supported if the server previously declared that it has the `parameterSubscribe` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"unsubscribeParameterUdpates"`
```
jhurliman, line 371:
```suggestion
- `op`: string `"unsubscribeParameterUpdates"`
```

jhurliman (approved):

docs/spec.md
```diff
@@ -132,6 +140,54 @@ Informs the client that channels are no longer available.
...
+  "parameters": [
+    { "name": "/int_param", "value": 2 },
+    { "name": "/float_param", "value": 1.2 },
+    { "name": "/string_param", "value": "foo" },
+    { "name": "/node/nested_ints_param", "value": [1, 2, 3] }
+  ]
+}
+```
+
+### Parameter Updates
```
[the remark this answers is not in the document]
achim-k: Yes, it has the same structure now. We can indeed remove it

docs/spec.md
```diff
@@ -235,6 +273,100 @@ Informs the client that channels are no longer available.
...
+#### Fields
+
+- `op`: string `"subscribeParameterUpdates"`
+- `parameterNames`: string[], leave empty to subscribe to all currently known parameters
+
+#### Example
+
+```json
+{
+  "op": "subscribeParameterUdpates",
```
jtbandes, line 337:
```suggestion
  "op": "subscribeParameterUpdates",
```

docs/spec.md
```diff
@@ -235,6 +273,100 @@ Informs the client that channels are no longer available.
...
+#### Fields
+
+- `op`: string `"unsubscribeParameterUpdates"`
+- `parameterNames`: string[], leave empty to unsubscribe from all parameter updates
+
+#### Example
+
+```json
+{
+  "op": "unsubscribeParameterUdpates",
```
jtbandes, line 360:
```suggestion
  "op": "unsubscribeParameterUpdates",
```

jtbandes (approved):

docs/spec.md
```diff
@@ -235,6 +273,100 @@ Informs the client that channels are no longer available.
...
+    { "name": "/float_param", "value": 4.1 }
+  ]
+}
+```
+
+### Subscribe Parameter Update
+
+Subscribe to parameter updates. Only supported if the server previously declared that it has the `parametersSubscribe` [capability](#server-info).
+
+Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUdpates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
```
achim-k, line 326:
```suggestion
Sending `subscribeParameterUpdates` multiple times will append the list of parameter subscriptions, not replace them. Note that parameters can be subscribed at most once. Hence, this operation will ignore parameters that are already subscribed. Use [unsubscribeParameterUpdates](#unsubscribe-parameter-update) to unsubscribe from existing parameter subscriptions.
```

achim-k:
@jhurliman @jtbandes I added 11b64ed to address https://github.com/foxglove/ws-protocol/pull/288#discussion_r1045876181. Could you PTAL again? Thanks!

docs/spec.md
```diff
@@ -132,6 +139,38 @@ Informs the client that channels are no longer available.
...
+#### Fields
+
+- `op`: string `"parameterValues"`
+- `parameters`: array of:
+  - `name`: string, name of the parameter
+  - `value`: number | boolean | string | number[] | boolean[] | string[]
+- `type`: string, one of
+  - `"serverInitiated"` - The message was sent by the server without being previously requested
+  - `"getParametersResponse"` - The message was sent as a response to the client's [Get Parameters request](#get-parameters)
+  - `"valueUpdate"` - The message was sent because one or more [subscribed](#subscribe-parameter-update) parameters have changed their value
```
jhurliman, line 155: I don't understand this field, what is a client supposed to do with it? The id field below already handles linking requests/responses together for getParameters.

jtbandes: I think this was answered above:

> For instance, if the client requests 2 parameters and the server replies with the corresponding 2 parameter values, the client should interpret this as the response to its previous request. It should not mean that there are only 2 parameters in total.

jhurliman: > the client should interpret this as the response to its previous request

Isn't that what the `id` field is for?

achim-k: The question is, do we have to distinguish between server initiated messages and value updates (parameter subscription callbacks)? If so, then we need this extra field.

achim-k: We currently allow the server to send parameters at random points of time (https://github.com/foxglove/ws-protocol/pull/288#discussion_r1038506796). The server could e.g. send all available parameters to newly connected clients. If that's really done is up to the server implementation.

jhurliman: What would the client code do?

```typescript
if (serverInitiated) {
 ... // ?
}
```

jhurliman: This makes sense, although I wouldn't call it "server initiated" since it has a more specific meaning of "this is the complete parameter list, delete all parameters not included in this message". I don't think supporting server-initiated full parameter lists is an important goal; I recommend removing the `type` field completely.

achim-k: OK, I will do that

achim-k, line 155:
```suggestion
```

docs/spec.md
```diff
@@ -132,6 +139,38 @@ Informs the client that channels are no longer available.
...
+- `id`: string | undefined. Only set when `type` is `"getParametersResponse"` and the [request's](#get-parameters) `id` field was set
```
achim-k, line 156:
```suggestion
- `id`: string | undefined. Only set when the [request's](#get-parameters) `id` field was set
```

docs/spec.md
```diff
@@ -132,6 +139,38 @@ Informs the client that channels are no longer available.
...
+```json
+{
+  "op": "parameterValues",
+  "parameters": [
+    { "name": "/int_param", "value": 2 },
+    { "name": "/float_param", "value": 1.2 },
+    { "name": "/string_param", "value": "foo" },
+    { "name": "/node/nested_ints_param", "value": [1, 2, 3] }
+  ],
+  "type": "getParametersResponse",
```
achim-k, line 169:
```suggestion
```

PR #298, opened 2022-12-12
Spec: Add `clockTopic` field to `serverInfo` message

achim-k:
**Public-Facing Changes**
- Update spec by adding `clockTopic` field to `serverInfo` message

**Description**
In ROS systems, the `/clock` topic tells subscribers about the current time of the system. Clients, like Foxglove Studio, may want to subscribe to this topic. In order to stay ROS agnostic, a `clockTopic` field was added which is used to inform clients about the clock topic name


achim-k:
@jhurliman @jtbandes PTAL

Relevant for https://github.com/foxglove/studio/issues/4908

defunctzombie:
I proposal that was discussed is to introduce a new opcode for "clock" and abstracting the topic subscription into the bridge implementation. Not sure if better or worse alternative than this approach.

jhurliman:
Closing in favor of https://github.com/foxglove/ws-protocol/pull/299

PR #299, opened 2022-12-13
Extend spec with publishing time data

achim-k:
**Public-Facing Changes**
- Allow server to publish time data (binary)
- Adapt typescript client/server accordingly

**Description**
- Extends the spec by adding publishing of time data under a new opcode (0x02). By publishing time data, clients can be informed about the progress of time even if no other message data is published.
- Updates `@foxglove/ws-protocol` accordingly

Supersedes #298

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -59,6 +59,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `clientPublish`: Allow clients to advertise channels to send data messages to the server
   - `parameters`: Allow clients to get & set parameters
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
+  - `time`: The server may publish binary [time](#time) messages
 
 #### Example
 
@@ -66,7 +67,7 @@ Each JSON message must be an object containing a field called `op` which identif
 {
   "op": "serverInfo",
   "name": "example server",
-  "capabilities": ["clientPublish"]
+  "capabilities": ["clientPublish", "time"]
 }
 ```
 
@@ -388,6 +389,7 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 
 - Inform clients about the latest server time. This allows accelerated, slowed, or stepped control over the progress of time.
 - If the server publishes time data, then timestamps of [published messages](#message-data) must originate from the same time source
+- The server may only publish time data if it previously declared support for it via the `time` [capability](#server-info)
 
 | Bytes | Type   | Description             |
 | ----- | ------ | ----------------------- |
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -28,6 +28,7 @@
 - [Advertise](#advertise) (json)
 - [Unadvertise](#unadvertise) (json)
 - [Message Data](#message-data) (binary)
+- [Time](#time) (binary)
 - [Parameter Values](#parameter-values) (json)
 
 ### Sent by client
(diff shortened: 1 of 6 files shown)


docs/spec.md
```diff
@@ -252,3 +253,14 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 | 4               | uint32  | subscription id                 |
 | 8               | uint64  | receive timestamp (nanoseconds) |
 | remaining bytes | uint8[] | message payload                 |
+
+### Time Data
+
+- Inform clients about the latest server time. This allows accelerated, slowed, or stepped control over the progress of time.
+- If the server publishes time data, then timestamps of [published messages](#message-data) must originate from the same time source
+- A final timestamp of 0 shall be published when the server decides to not publish time data anymore.
```
achim-k, line 261:
```suggestion
```

docs/spec.md
```diff
@@ -28,6 +28,7 @@
 - [Advertise](#advertise) (json)
 - [Unadvertise](#unadvertise) (json)
 - [Message Data](#message-data) (binary)
+- [Time Data](#time-data) (binary)
```
achim-k, line 31:
```suggestion
- [Time](#time) (binary)
```

docs/spec.md
```diff
@@ -252,3 +253,13 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 | 4               | uint32  | subscription id                 |
 | 8               | uint64  | receive timestamp (nanoseconds) |
 | remaining bytes | uint8[] | message payload                 |
+
+### Time Data
```
jtbandes: naming: Time or Time Update?

jtbandes (approved):

achim-k:
@jtbandes @jhurliman I have now also added the necessary adaptation to `@foxglove/ws-protocol` to this PR. Let me know if I should rather do open a new PR for that

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -124,6 +124,15 @@ export default class FoxgloveServer {
     }
   }
 
+  /**
+   * Emit a time udpate to clients.
```
jtbandes, line 128:
```suggestion
   * Emit a time update to clients.
```

typescript/ws-protocol/src/types.ts
```diff
@@ -78,12 +80,22 @@ export type MessageData = {
   timestamp: bigint;
   data: DataView;
 };
+export type TimeData = {
```
jtbandes, line 83: name doesn't match opcode – was there a naming conflict?

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -375,4 +384,21 @@ export default class FoxgloveServer {
...
+    // attempt to detect support for {fin: false}
+    if (connection.send.length > 1) {
+      connection.send(msg, { fin: true });
+    } else if (typeof Blob === "function") {
+      connection.send(new Blob([msg]));
+    } else {
+      const buffer = new Uint8Array(msg.byteLength);
+      buffer.set(new Uint8Array(msg.buffer), 0);
+      connection.send(buffer);
+    }
```
jtbandes, line 402: `fin:false` is not useful here, since we already have all the data in a single buffer, you should be able to use the same API for both web and node and skip the Blob stuff

jtbandes (commented):
I don't mind having it in the same PR once we are aligned on the semantics.

typescript/ws-protocol/src/FoxgloveServer.ts
achim-k: seems to be a common mistake of mine :sweat_smile:

typescript/ws-protocol/src/types.ts
achim-k: nope, will update it

jhurliman (approved):

PR #319, opened 2023-01-05
Add parameter support to the typescript implementation

achim-k:
**Public-Facing Changes**
- Add parameter support to the typescript implementation

**Description**
Adds support for parameters (see #288) to the typescript package

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveServer.ts
+++ b/typescript/ws-protocol/src/FoxgloveServer.ts
@@ -26,7 +26,7 @@ type ClientInfo = {
   parameterSubscriptions: Set<string>;
 };
 
-export type SingleClient = { client: ClientInfo };
+type SingleClient = { client: ClientInfo };
 
 type EventTypes = {
   error: (error: Error) => void;
@@ -42,7 +42,10 @@ type EventTypes = {
   /** A client stopped advertising a channel. */
   unadvertise: (channel: { channelId: ChannelId } & SingleClient) => void;
   /** Request to retrieve parameter values has been received. */
-  getParameters: (request: { parameterNames: string[]; id?: string } & SingleClient) => void;
+  getParameters: (
+    request: { parameterNames: string[]; id?: string },
+    clientConnection: IWebSocket | undefined,
+  ) => void;
   /** Request to set parameter values has been received. */
   setParameters: (parameters: Parameter[]) => void;
   /** Request to subscribe to parameter value updates has been received. */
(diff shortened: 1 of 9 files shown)


jhurliman (approved):

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -154,13 +193,19 @@ export default class FoxgloveServer {
...
     this.send(connection, {
       op: "serverInfo",
       name: this.name,
-      capabilities: [ServerCapability.clientPublish, ServerCapability.time],
+      capabilities: [
+        ServerCapability.clientPublish,
+        ServerCapability.time,
+        ServerCapability.parameters,
+        ServerCapability.parametersSubscribe,
+      ],
```
jtbandes, line 208: Ideally we would not always send these capabilities, they would be up to the developer using FoxgloveServer to enable or disable them. If someone wants to use this class but doesn't want to offer parameter support, they should be able to suppress the capability from being sent.

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -39,6 +41,14 @@ type EventTypes = {
   advertise: (channel: ClientChannel & SingleClient) => void;
   /** A client stopped advertising a channel. */
   unadvertise: (channel: { channelId: ChannelId } & SingleClient) => void;
+  /** Request to retrieve parameter values has been received. */
+  getParameters: (request: { parameterNames: string[]; id?: string } & SingleClient) => void;
```
jtbandes, line 45: Could simplify this by making the client a 2nd argument of the event handler method rather than merging it with the event message object itself. I guess we already have this pattern for a few other events though.

jtbandes (approved):
Nice tests 👍 

It would also be helpful to add support for parameters to the existing example servers under `ws-protocol-examples`, or add a new example server/client to demonstrate params.

typescript/ws-protocol/src/FoxgloveServer.ts
achim-k: I agree. I will refactor it once this PR is in

typescript/ws-protocol/src/FoxgloveServer.ts
achim-k: > Could simplify this by making the client a 2nd argument of the event handler method rather than merging it with the event message object itself.

I ended up doing this. And instead of passing the entire client info only the `IWebsocket` instance is passed, which is already exported in `types`

jtbandes (approved):

PR #320, opened 2023-01-06
Allow to configure capabilities in typescript server implementation

achim-k:
**Public-Facing Changes**
- Allow to configure capabilities in typescript server implementation

**Description**
So far, capabilities of the typescript server implementation were hardcoded. With this PR, capabilities can be configured via a constructor argument. See also https://github.com/foxglove/ws-protocol/pull/319#discussion_r1062854374

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveServer.ts
+++ b/typescript/ws-protocol/src/FoxgloveServer.ts
@@ -56,17 +56,33 @@ type EventTypes = {
 
 const log = createDebug("foxglove:server");
 
+const REQUIRED_CAPABILITY_BY_OPERATION: Record<
+  ClientMessage["op"],
+  keyof typeof ServerCapability | undefined
+> = {
+  subscribe: undefined,
+  unsubscribe: undefined,
+  advertise: ServerCapability.clientPublish,
+  unadvertise: ServerCapability.clientPublish,
+  getParameters: ServerCapability.parameters,
+  setParameters: ServerCapability.parameters,
+  subscribeParameterUpdates: ServerCapability.parametersSubscribe,
+  unsubscribeParameterUpdates: ServerCapability.parametersSubscribe,
+};
+
 export default class FoxgloveServer {
   static SUPPORTED_SUBPROTOCOL = "foxglove.websocket.v1";
 
   readonly name: string;
+  readonly capabilities: string[];
   private emitter = new EventEmitter<EventTypes>();
   private clients = new Map<IWebSocket, ClientInfo>();
   private nextChannelId: ChannelId = 0;
   private channels = new Map<ChannelId, Channel>();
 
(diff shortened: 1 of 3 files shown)


typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -284,6 +295,16 @@ export default class FoxgloveServer {
   }
 
   private handleClientMessage(client: ClientInfo, message: ClientMessage): void {
+    const requiredCapability = REQUIRED_CAPABILITY_BY_OPERATION[message.op];
```
jtbandes, line 298: Should we have a similar check for when the server tries to **send** parameter values or time?

achim-k: I will add that

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -163,6 +163,14 @@ export default class FoxgloveServer {
    * Emit a time update to clients.
    */
   broadcastTime(timestamp: bigint): void {
+    if (!this.capabilities.includes(ServerCapability.time)) {
+      log(
+        "Sending time data is only supported if the server has declared the '%s' capability before.",
```
jtbandes, line 168:
```suggestion
        "Sending time data is only supported if the server has declared the '%s' capability.",
```

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -175,6 +183,14 @@ export default class FoxgloveServer {
    * @param connection Optional connection when parameter values are to be sent to a single client
    */
   publishParameterValues(parameters: Parameter[], id?: string, connection?: IWebSocket): void {
+    if (!this.capabilities.includes(ServerCapability.parameters)) {
+      log(
+        "Publishing parameter values is only supported if the server has declared the '%s' capability before.",
```
jtbandes, line 188:
```suggestion
        "Publishing parameter values is only supported if the server has declared the '%s' capability.",
```

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -189,6 +205,14 @@ export default class FoxgloveServer {
    * @param parameters Parameter values
    */
   updateParameterValues(parameters: Parameter[]): void {
+    if (!this.capabilities.includes(ServerCapability.parametersSubscribe)) {
+      log(
+        "Publishing parameter value updates is only supported if the server has declared the '%s' capability before.",
```
jtbandes, line 210:
```suggestion
        "Publishing parameter value updates is only supported if the server has declared the '%s' capability.",
```

jtbandes (approved):

PR #328, opened 2023-01-09
Allow specifying a request id when setting parameters

achim-k:
**Public-Facing Changes**
- Allow specifying a request id in the `setParameters` operation

**Description**
The server may reject requests to update parameters for various reasons such as e.g. parameters being readonly. To allow a client to verify if parameter updates were successful, this PR adds an optional request ID field to the `setParameters` operation.


jhurliman (approved):
I'm not sure about the choice of string vs number, but I don't have a strong opinion either way.

PR #334, opened 2023-01-11
Spec: Allow to send additional metadata to clients

achim-k:
**Public-Facing Changes**
- Allow to send additional server options to clients

**Description**
Allows to inform clients about additional server options. E.g. for [foxglove-bridge  ](https://github.com/foxglove/ros-foxglove-bridge) we would like to communicate the ROS_DISTRO to the client

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -61,6 +61,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
   - `time`: The server may publish binary [time](#time) messages
 - `supportedEncodings`: array of strings | informing the client about which encodings may be used for client-side publishing. Only set if client publishing is supported.
+- `options`: optional map of key-value pairs
 
 #### Example
 
@@ -69,7 +70,10 @@ Each JSON message must be an object containing a field called `op` which identif
   "op": "serverInfo",
   "name": "example server",
   "capabilities": ["clientPublish", "time"],
-  "supportedEncodings": ["json"]
+  "supportedEncodings": ["json"],
+  "options": {
+    "key": "value"
+  }
 }
 ```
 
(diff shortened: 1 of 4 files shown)


docs/spec.md
```diff
@@ -61,6 +61,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
   - `time`: The server may publish binary [time](#time) messages
 - `supportedEncodings`: array of strings | informing the client about which encodings may be used for client-side publishing. Only set if client publishing is supported.
+- `options`: optional map of key-value pairs
```
jtbandes, line 64: Hm, maybe call this `metadata` to mirror mcap channel metadata? What do you think @jhurliman?  cc @james-rms

achim-k: I agree, it sounds better

```suggestion
- `metadata`: optional map of key-value pairs
```

docs/spec.md
```diff
@@ -69,7 +70,10 @@ Each JSON message must be an object containing a field called `op` which identif
   "op": "serverInfo",
   "name": "example server",
   "capabilities": ["clientPublish", "time"],
-  "supportedEncodings": ["json"]
+  "supportedEncodings": ["json"],
+  "options": {
```
achim-k, line 74:
```suggestion
  "metadata": {
```

typescript/ws-protocol/src/types.ts
```diff
@@ -71,6 +71,7 @@ export type ServerInfo = {
   name: string;
   capabilities: string[];
   supportedEncodings?: string[];
+  options?: Record<string, unknown>;
```
achim-k, line 74:
```suggestion
  metadata?: Record<string, unknown>;
```

jhurliman (approved):

typescript/ws-protocol/src/types.ts
```diff
@@ -71,6 +71,7 @@ export type ServerInfo = {
   name: string;
   capabilities: string[];
   supportedEncodings?: string[];
+  metadata?: Record<string, unknown>;
```
jhurliman, line 74: `unknown` is fine since we only support JSON serialization. If we used a different serialization, this might need to be `<string, string>` instead. I think we can leave it how it is.

achim-k:
```suggestion
  metadata?: Record<string, string>;
```

PR #344, opened 2023-01-17
Extend spec with services capability

achim-k:
**Public-Facing Changes**
- Adds additional operations to allow clients to call services that are advertised by the server

**Description**
Adds additional operations to allow clients to call services that are advertised by the server

FG-1456

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -30,6 +30,9 @@
 - [Message Data](#message-data) (binary)
 - [Time](#time) (binary)
 - [Parameter Values](#parameter-values) (json)
+- [AdvertiseServices](#advertise-services) (json)
+- [UnadvertiseServices](#unadvertise-services) (json)
+- [Service Call Response](#service-call-response) (binary)
 
 ### Sent by client
 
@@ -42,6 +45,7 @@
 - [Set Parameters](#set-parameters) (json)
 - [Subscribe Parameter Update](#subscribe-parameter-update) (json)
 - [Unsubscribe Parameter Update](#unsubscribe-parameter-update) (json)
+- [Service Call Request](#service-call-request) (binary)
 
 ## JSON messages
 
@@ -60,6 +64,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `parameters`: Allow clients to get & set parameters
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
   - `time`: The server may publish binary [time](#time) messages
+  - `services`: Allow clients to call services
 - `supportedEncodings`: array of strings | informing the client about which encodings may be used for client-side publishing. Only set if client publishing is supported.
 - `metadata`: optional map of key-value pairs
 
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -176,6 +181,60 @@ Informs the client about parameters. Only supported if the server declares the `
...
+
+- `op`: string `"advertiseServices"`
+- `services`: array of:
+  - `id`: number. The server may reuse ids when services disappear and reappear, but only if the services keeps the exact same name, type, and schema. Clients will use this unique id to cache schema info and deserialization routines.
+  - `name`: string
+  - `type`: string
+  - `schema`: string
+  - `acceptedEncodings`: array of
+    - `id`: number
+    - `name`: string, name of an encoding which can be used to encode service call requests
```
achim-k, line 198: We could also just make that a string and allow only a single request encoding?

docs/spec.md
```diff
@@ -60,6 +64,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `parameters`: Allow clients to get & set parameters
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
   - `time`: The server may publish binary [time](#time) messages
+  - `services`: Allow clients to call services
```
jhurliman, line 67: Alternatively, if the server never advertises a service there are no services for the client to call :)

docs/spec.md
jhurliman: I like the array, but why the `id` field?

docs/spec.md
```diff
@@ -405,3 +464,26 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+- Request to call a service that has been advertised by the server.
+- Only supported if the server previously declared the `services` [capability](#server-info).
+
+| Bytes           | Type    | Description                                                                                  |
+| --------------- | ------- | -------------------------------------------------------------------------------------------- |
+| 1               | opcode  | 0x02                                                                                         |
+| 4               | uint32  | service id                                                                                   |
+| 8               | uint64  | call id, a unique number to identify the corresponding service response                      |
+| 4               | uint32  | id of the request encoding, as specified in the [service advertisement](#advertise-services) |
```
jhurliman, line 478: The id above makes more sense now, to avoid a length-prefixed string here. I have a preference for a length-prefixed string instead of introducing the `encoding_id` concept.

jhurliman (approved):

docs/spec.md
achim-k: What length would you suggest? Or you mean to have a second field with the length of the encoding string?

Alternatively, we could make `acceptedEncoding` a string array and refer to the index here?

docs/spec.md
```diff
@@ -405,3 +464,26 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+### Service Call Response
+
+- Provides the response to a previous [service call](#service-call-request).
+
+| Bytes           | Type    | Description      |
+| --------------- | ------- | ---------------- |
+| 1               | opcode  | 0x03             |
+| 8               | uint64  | call id          |
+| remaining bytes | uint8[] | response payload |
```
achim-k, line 489: mention that the encoding matches the request encoding

docs/spec.md
```diff
@@ -30,6 +30,9 @@
 - [Message Data](#message-data) (binary)
 - [Time](#time) (binary)
 - [Parameter Values](#parameter-values) (json)
+- [AdvertiseServices](#advertise-services) (json)
+- [UnadvertiseServices](#unadvertise-services) (json)
```
jtbandes, line 34: nit
```suggestion
- [Advertise Services](#advertise-services) (json)
- [Unadvertise Services](#unadvertise-services) (json)
```

docs/spec.md
```diff
@@ -176,6 +181,57 @@ Informs the client about parameters. Only supported if the server declares the `
...
+#### Fields
+
+- `op`: string `"advertiseServices"`
+- `services`: array of:
+  - `id`: number. The server may reuse ids when services disappear and reappear, but only if the services keeps the exact same name, type, and schema. Clients will use this unique id to cache schema info and deserialization routines.
+  - `name`: string
+  - `type`: string
+  - `requestSchema`: string
+  - `responseSchema`: string
+  - `acceptedEncodings`: array of string, encodings which can be used to encode service call requests
```
jtbandes, line 197: does accepted encodings need to be on individual service level or should it just be for the whole server?

docs/spec.md
achim-k:
```suggestion
  - `services`: Server can advertise services which clients may call
```

jhurliman: This capability seems redundant. If a server can't advertise services, then it won't advertise any services. If it does advertise a service, why should a client make an additional check to ensure the `services` capability was set? And what should it do if a service is advertised, but the capability is not?

jhurliman (approved):

docs/spec.md
```diff
@@ -176,6 +181,55 @@ Informs the client about parameters. Only supported if the server declares the `
...
+### Advertise Services
+
+Informs the client about available services. Only supported if the server declares the `services` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"advertiseServices"`
+- `services`: array of:
+  - `id`: number. The server may reuse ids when services disappear and reappear, but only if the services keeps the exact same name, type, and schema. Clients will use this unique id to cache schema info and deserialization routines.
+  - `name`: string
```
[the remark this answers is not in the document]
achim-k: It may happen that a service is unadvertised and a new service with the same name but different type is advertised. In this case the name is not unique.

docs/spec.md
achim-k: It may help a client to know whether the server will potentially announce services or not. But I do not have a strong opinion on this. @jtbandes what do you think?

PR #347, opened 2023-01-27
@foxglove/ws-protocol: Add services capability, minor cleanup

achim-k:
**Public-Facing Changes**
- Updates `@foxglove/ws-protocol` to implement service operations added in #344 

**Description**
- Updates `@foxglove/ws-protocol` to implement service operations added in #344 
- Minor cleanup in the way client messages are being handled

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveClient.ts
+++ b/typescript/ws-protocol/src/FoxgloveClient.ts
@@ -13,8 +13,12 @@ import {
   Parameter,
   ParameterValues,
   ServerMessage,
+  ServiceCallResponse,
+  ServiceId,
+  Service,
   SubscriptionId,
   Time,
+  ServiceCallRequest,
 } from "./types";
 
 type EventTypes = {
@@ -28,7 +32,10 @@ type EventTypes = {
   time: (event: Time) => void;
   advertise: (newChannels: Channel[]) => void;
   unadvertise: (removedChannels: ChannelId[]) => void;
+  advertiseServices: (newServices: Service[]) => void;
+  unadvertiseServices: (removedServices: ServiceId[]) => void;
   parameterValues: (event: ParameterValues) => void;
+  serviceCallResponse: (event: ServiceCallResponse) => void;
 };
 
 export default class FoxgloveClient {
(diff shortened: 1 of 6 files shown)


jhurliman (approved):

typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -173,6 +192,31 @@ export default class FoxgloveClient {
...
+    view.setUint32(offset, request.encoding.length, true);
+    offset += 4;
+    payload.set(encoding, offset);
+    offset += encoding.length;
+    const data = new Uint8Array(
+      request.data.buffer,
+      request.data.byteOffset,
+      request.data.byteLength,
+    );
+    payload.set(data, offset);
```
achim-k, line 216: @jhurliman is there a better way to do that? I feel like we have to do to many Uint8Array / Dataview conversions

typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -173,6 +192,31 @@ export default class FoxgloveClient {
     this.ws.send(payload);
   }
 
+  sendServiceRequest(request: ServiceCallRequest): void {
```
achim-k, line 195: I think this is more user friendly as otherwise the user has to fill in the correct op

```suggestion
  sendServiceRequest(request: Omit<ServiceCallRequest, "op">): void {
```

achim-k: Can also make `ServiceCallRequest` and `ServiceCallResponse` to share the base properties then

typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -173,6 +192,31 @@ export default class FoxgloveClient {
...
+    const utf8Encode = new TextEncoder();
```
jhurliman, line 196: This should be a global constant, don't initialize a TextEncoder every call

typescript/ws-protocol/src/FoxgloveClient.ts
jhurliman: This all looks right, unless `request.data` is already a Uint8Array. These conversions are lightweight

PR #352, opened 2023-02-02
Example server implementation: Disallow multiple client subscriptions per channel per client

### LINKED ISSUE #233: Spec disallows multiple subscriptions per channel per client, but reference implementations allow it
The spec says:

> A client may only have one subscription for each channel at a time.

However, our reference implementations allow this by keeping a mapping from channel id to multiple subscription ids for each client:

https://github.com/foxglove/ws-protocol/blob/c883a5f7b9a23f87b6131b8d7f4f162567b3e8eb/typescript/ws-protocol/src/FoxgloveServer.ts#L24

https://github.com/foxglove/ws-protocol/blob/c883a5f7b9a23f87b6131b8d7f4f162567b3e8eb/cpp/foxglove-websocket/include/foxglove/websocket/server.hpp#L106 

https://github.com/foxglove/ws-protocol/blob/c883a5f7b9a23f87b6131b8d7f4f162567b3e8eb/python/src/foxglove_websocket/server/client_state.py#L23-L25

achim-k:
**Public-Facing Changes**
- Python / Typescript example implementation: Disallow multiple subscriptions per channel per client

**Description**
The spec says that clients may only have one subscription for each channel at a time. However, our reference implementation allowed having more than one subscription. This PR fixes this for the following reference implementations:

- [x] Typescript
- [x] Python
- [x] C++  (Fixed in #375)

fixes #233
fixes FG-957

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveServer.ts
+++ b/typescript/ws-protocol/src/FoxgloveServer.ts
@@ -26,7 +26,7 @@ type ClientInfo = {
   name: string;
   connection: IWebSocket;
   subscriptions: Map<SubscriptionId, ChannelId>;
-  subscriptionsByChannel: Map<ChannelId, Set<SubscriptionId>>;
+  subscriptionsByChannel: Map<ChannelId, SubscriptionId>;
   advertisements: Map<ClientChannelId, ClientChannel>;
   parameterSubscriptions: Set<string>;
 };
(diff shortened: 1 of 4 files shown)


typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -415,12 +411,7 @@ export default class FoxgloveServer {
           log("client %s subscribed to channel %d", client.name, channelId);
           const firstSubscription = !this.anySubscribed(channelId);
           client.subscriptions.set(subId, channelId);
-          let subs = client.subscriptionsByChannel.get(channelId);
-          if (!subs) {
-            subs = new Set();
-            client.subscriptionsByChannel.set(channelId, subs);
-          }
-          subs.add(subId);
+          client.subscriptionsByChannel.set(channelId, subId);
```
jtbandes, line 414: Error/warning message for already subscribed?

jtbandes (approved):

PR #354, opened 2023-02-03
Allow unsetting (removing) parameters

### LINKED ISSUE #346: Allow unsetting of parameters
**Description**
The spec currently does not support unsetting of parameters. This request came up here: https://github.com/foxglove/studio/pull/5227#discussion_r1087356472

achim-k:
**Public-Facing Changes**
- Allow unsetting (removing) parameters

**Description**
See https://github.com/foxglove/studio/pull/5227#discussion_r1087356472 

Fixes #346

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -367,7 +367,7 @@ Set one or more parameters. Only supported if the server previously declared tha
 - `op`: string `"setParameters"`
 - `parameters`: array of:
   - `name`: string
-  - `value`: number | boolean | string | number[] | boolean[] | string[]
+  - `value`: number | boolean | string | number[] | boolean[] | string[] | undefined. When set to `undefined`, the parameter shall be unset (removed).
 - `id`: string | undefined, arbitrary string used for identifying the corresponding server [response](#parameter-values). If this field is not set, the server may not send a response to the client.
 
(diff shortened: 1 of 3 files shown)


jhurliman (approved):

docs/spec.md
```diff
@@ -367,7 +367,7 @@ Set one or more parameters. Only supported if the server previously declared tha
 - `op`: string `"setParameters"`
 - `parameters`: array of:
   - `name`: string
-  - `value`: number | boolean | string | number[] | boolean[] | string[]
+  - `value`: number | boolean | string | number[] | boolean[] | string[] | undefined. When set to `undefined`, the parameter shall be unset (removed).
```
achim-k, line 370: IMO, we should use `null` instead of `undefined` as it is much more explicit that the value is being unset. @jhurliman / @jtbandes what do you think?

jtbandes: I think it's dumb that JS has both `null` and `undefined`. Usually I choose to believe I live in a world where `null` does not exist. However, unfortunately JSON only supports `null` and not `undefined`. So actually this spec does not exactly make sense as written... We either need to use null or say that the `value` key can be absent from the JSON.

PR #363, opened 2023-02-09
Mention base64 encoding in protocol specification

jkuszmaul:
Until I actually encountered errors in the code, I hadn't realized that base64 encoding was needed, partially because I don't use things that actually enforce unicode strings all that often. Since the string "base64" wasn't mentioned anywhere in the docs, at least bring it up to remind people that this is relevant.

I imagine other people might have similar questions.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -51,6 +51,9 @@
 
 Each JSON message must be an object containing a field called `op` which identifies the type of message.
 
+Additionally, each JSON message must encode any binary data fields (e.g.,
+protobuf or flatbuffer schemas) with base64.
+
 ### Server Info
 
 - This message is always sent to new clients upon connection.


docs/spec.md
```diff
@@ -51,6 +51,9 @@
 
 Each JSON message must be an object containing a field called `op` which identifies the type of message.
 
+Additionally, each JSON message must encode any binary data fields (e.g.,
+protobuf or flatbuffer schemas) with base64.
```
jtbandes, line 55: 🤷  Technically, this isn't a requirement of the websocket protocol. Someone could choose to represent binary data with base36 instead of base64. The only requirement of websockets in general is that text messages are valid UTF8, and we're additionally adding a requirement that they should be valid JSON. Maybe it would be more accurate to mention that Foxglove Studio expects the schema field to be base64-encoded if the schemas are represented with binary data, and link to [these cods](https://foxglove.dev/docs/studio/connection/custom)? But I see that below we already link to this on the docs specific to the `schema` field...

jhurliman: We should call out the well-known binary schemas by name, so people scanning the docs can quickly see "I'm using protobufs, oh I need to base64-encode my FileDescriptorSet" or "I'm using ROS, I can send my message definitions directly as UTF8".

jtbandes: You think that should live in the actual ws-protocol spec, or in an appendix or in studio docs?

jhurliman: Thinking through this, maybe we should make base64 encoding of binary data in JSON messages part of the spec. I don't think it harms any use cases (i.e. someone wanting to use base36 encoding instead and expecting Studio or other clients to support that), and it clarifies the implementation details.

defunctzombie:
@jtbandes @jhurliman what's the latest thinking here? Do we want to include this or do we want to say this is outside of the core spec. Do we want to add an appendix where we cover various supported items.

Practically what is supported by this protocol is specific to what studio supports since this protocol exists to bridge systems to Studio.

jtbandes:
I'm going to go ahead and close this PR since I believe we already state the relevant base64-encoding requirements on [this docs page](https://foxglove.dev/docs/studio/connection/custom), which is linked from the `schema` field description in the spec.

PR #368, opened 2023-02-14
spec: Add connection graph capability

achim-k:
**Public-Facing Changes**
- Adds a new capability to subscribe to connection graph updates which includes information about topic publishers/subscribers as well as service providers

**Description**
See above

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -33,6 +33,7 @@
 - [Advertise Services](#advertise-services) (json)
 - [Unadvertise Services](#unadvertise-services) (json)
 - [Service Call Response](#service-call-response) (binary)
+- [Connection Graph Update](#connection-graph-update) (json)
 
 ### Sent by client
 
@@ -46,6 +47,8 @@
 - [Subscribe Parameter Update](#subscribe-parameter-update) (json)
 - [Unsubscribe Parameter Update](#unsubscribe-parameter-update) (json)
 - [Service Call Request](#service-call-request) (binary)
+- [Subscribe connection graph updates](#subscribe-connection-graph-updates) (json)
+- [Unubscribe connection graph updates](#unsubscribe-connection-graph-updates) (json)
 
 ## JSON messages
 
@@ -65,6 +68,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `parametersSubscribe`: Allow clients to subscribe to parameter changes
   - `time`: The server may publish binary [time](#time) messages
   - `services`: Allow clients to call services
+  - `connectionGraph`: Allow clients to subscribe to updates to the connection graph
 - `supportedEncodings`: array of strings | informing the client about which encodings may be used for client-side publishing or service call requests/responses. Only set if client publishing or services are supported.
 - `metadata`: optional map of key-value pairs
 - `sessionId`: optional string. Allows the client to understand if the connection is a re-connection or if it is connecting to a new server instance. This can for example be a timestamp or a UUID.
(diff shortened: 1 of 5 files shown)


docs/spec.md
```diff
@@ -433,6 +473,38 @@ Unsubscribe from parameter updates. Only supported if the server previously decl
 }
 ```
 
+### Subscribe connection graph updates
```
achim-k, line 476: @jtbandes had a better idea to add an operation that allows to subscribe/unsubscribe to/from features. I just couldn't come up with a good name for this operation. Any proposals?

docs/spec.md
```diff
@@ -232,6 +236,42 @@ Informs the client about services that are no longer available. Only supported i
...
+#### Fields
+
+- `op`: string `"connectionGraphUpdate"`
+- `topicUpdates`: array of newly added or updated topics
+  - `name`: string, topic name
+  - `publisherIds`: array of string, list of publisher IDs
+  - `subscriberIds`: array of string, list of subscriber IDs
+- `serviceUpdates`: array of newly added or updated services
+  - `name`: string, service name
+  - `providerIds`: array of string, list of service provider IDs
```
achim-k, line 252: I don't know if we should include this here, as services can in the end have only one provider...?

docs/spec.md
```diff
@@ -232,6 +236,42 @@ Informs the client about services that are no longer available. Only supported i
...
+  "topicUpdates": [
+    {
+      "name": "/foo",
+      "publisherIds": ["/node_1"],
+      "subscriberIds": ["/node_2"]
+    },
+    { "name": "/bar", "publisherIds": ["/node_1"], "subscriberIds": [] }
+  ],
+  "serviceUpdates": [{ "name": "/set_bool", "providerIds": ["/node_2"] }],
+  "removedTopics": ["/baz"],
```
achim-k, line 270: This could also be communicated by a topic with zero publishers and zero subscribers

docs/spec.md
jhurliman: For ROS that's true. I say leave it to support non-ROS systems with multiple providers for a single service endpoint.

docs/spec.md
jhurliman: It may be valuable to communicate the presence of a topic or service that has no publishers or subscribers (yet). I'm not sure if this is possible in ROS, but you could imagine a robot framework that uses a static configuration of topics before any nodes register as publishers or subscribers.

docs/spec.md
jhurliman: Maybe

```json
{
  "op": "subscribeFeature",
  "feature": "connectionGraphUpdate"
}
```

But I think this non-generic version is fine too.

jhurliman (approved):

docs/spec.md
```diff
@@ -419,6 +458,38 @@ Unsubscribe from parameter updates. Only supported if the server previously decl
 }
 ```
 
+### Subscribe connection graph updates
+
+Subscribe to [connection graph updates](#connection-graph-update). Only supported if the server previously declared that it has the `connectionGraph` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"subscribeConnectionGraphUpdate"`
```
jtbandes, line 467: naming nit: not sure the "Update" part is very useful, we could just call it `(un)subscribeConnectionGraph`

docs/spec.md
```diff
@@ -232,6 +236,41 @@ Informs the client about services that are no longer available. Only supported i
...
+  - `name`: topic name
+  - `publisherIds`: array of string, list of publisher IDs
+- `subscribedTopics`: array of
+  - `name`: topic name
+  - `subscriberIds`: array of string, list of subscriber IDs
+- `advertisedServices`: array of
+  - `name`: service name
+  - `providerIds`: array of string, list of provider IDs
+- `removedTopics`: array of string, names of topics that have been removed
+- `removedServices`: array of string, names of services that have been removed
```
jtbandes, line 256: Need to clarify the semantics of these – do they replace all the data from the previous message? Do they only replace on an individual topic level? Or are they purely additive?

jhurliman (approved):

jtbandes (approved):
🚀

PR #374, opened 2023-02-22
Force installation of conan 1.x

achim-k:
**Public-Facing Changes**
None

**Description**
Fixes cpp CI failures such as in https://github.com/foxglove/ws-protocol/actions/runs/4243120802/jobs/7375459524

### DIFF
--- a/cpp/dev.Dockerfile
+++ b/cpp/dev.Dockerfile
@@ -22,7 +22,7 @@ ENV CXX=clang++
 WORKDIR /src
 
 FROM base as build
-RUN pip --no-cache-dir install conan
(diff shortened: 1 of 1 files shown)


jhurliman (approved):

jtbandes:
On the other hand, should we consider using 2.x? Is that the latest version?

achim-k:
> On the other hand, should we consider using 2.x? Is that the latest version?

We could do that. I'm currently working on copying the c++ server implementation from foxglove_bridge, so have to get familiar with conan anyway. I'll look into conan 2 then.  Let's leave this PR on hold, will probably be obsolete later

achim-k:
Superseded by #375

PR #386, opened 2023-03-08
Achim/update cpp implementation

achim-k:
### Public-Facing Changes
- Declares a dependency on asio (fixes CI build errors)
- Updates the cpp implementation to match the latest foxglove_bridge version

### Description
Not sure why the cpp builds were working before. In https://docs.websocketpp.org/faq.html it says:

> **Can WebSocket++ be used with standalone Asio**
>
> Yes. The process is the same as used with standalone Asio itself. Define ASIO_STANDALONE before including Asio or WebSocket++ headers. You will need to download a copy of the Asio headers separately (http://www.think-async.com/) and make sure they are in your build system's include path.

@jtbandes, I haven't yet created a 1.0.0 recipe on conan-center. If OK, I would just update the 1.0.0 tag after this PR got merged

### DIFF
--- a/cpp/foxglove-websocket/include/foxglove/websocket/common.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/common.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <array>
 #include <cstring>
 #include <stdint.h>
 #include <string>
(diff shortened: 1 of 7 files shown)


cpp/foxglove-websocket/include/foxglove/websocket/common.hpp
```diff
@@ -13,6 +14,12 @@ constexpr char CAPABILITY_TIME[] = "time";
 constexpr char CAPABILITY_PARAMETERS[] = "parameters";
 constexpr char CAPABILITY_PARAMETERS_SUBSCRIBE[] = "parametersSubscribe";
 constexpr char CAPABILITY_SERVICES[] = "services";
+constexpr char CAPABILITY_CONNECTION_GRAPH[] = "connectionGraph";
+
+constexpr std::array<const char*, 5> DEFAULT_CAPABILITIES = {
```
jtbandes, line 19: I think we should be able to omit this and not have to worry about getting the number right?
```suggestion
constexpr std::array DEFAULT_CAPABILITIES = {
```

jtbandes (approved):

cpp/foxglove-websocket/include/foxglove/websocket/common.hpp
achim-k: You are right, this is much better

PR #387, opened 2023-03-08
Bump @foxglove/ws-protocol version

achim-k:
### Public-Facing Changes
None

### Description
- Bumps @foxglove/ws-protocol
- Updates version used in @foxglove/ws-protocol-examples

### DIFF
--- a/typescript/ws-protocol-examples/package.json
+++ b/typescript/ws-protocol-examples/package.json
@@ -67,7 +67,7 @@
     "@foxglove/schemas": "^1.0.0",
     "@foxglove/wasm-lz4": "^1.0.2",
     "@foxglove/wasm-zstd": "^1.0.1",
-    "@foxglove/ws-protocol": "^0.4.0",
+    "@foxglove/ws-protocol": "^0.5.0",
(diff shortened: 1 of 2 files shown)


typescript/ws-protocol-examples/package.json
```diff
@@ -67,7 +67,7 @@
     "@foxglove/schemas": "^1.0.0",
     "@foxglove/wasm-lz4": "^1.0.2",
     "@foxglove/wasm-zstd": "^1.0.1",
-    "@foxglove/ws-protocol": "^0.4.0",
+    "@foxglove/ws-protocol": "^0.5.0",
```
jtbandes, line 70: Would we want to bump this to 0.5.1 in a separate PR? I suppose we will also want to bump the version of the examples package and publish that too.

jtbandes (approved):

typescript/ws-protocol-examples/package.json
achim-k: I can also do it in this PR if you don't mind?

PR #389, opened 2023-03-10
cpp: Explicitly enable cpp17

achim-k:
### Public-Facing Changes
None

### Description
I though that conan would automaticlaly do this, but apparently this is not the case: https://github.com/conan-io/conan-center-index/pull/16500#issuecomment-1463916745
Anyway, it's still a good practice to enable it explicitly in case it is build without conan


jtbandes (approved):
In the mcap repo we use `set(CMAKE_CXX_STANDARD 17)`. Is there a difference?

achim-k:
> In the mcap repo we use `set(CMAKE_CXX_STANDARD 17)`. Is there a difference?

It is basically the same, but `set_target_properties` allows for a more fine-grained control

PR #430, opened 2023-04-27
Add mcap-record example

jtbandes:
### Public-Facing Changes

Added `npx @foxglove/ws-protocol-examples@latest mcap-record` which connects to a WebSocket server and records data into an MCAP file.

### DIFF
--- a/typescript/ws-protocol-examples/src/examples/mcap-record.ts
+++ b/typescript/ws-protocol-examples/src/examples/mcap-record.ts
@@ -56,6 +56,7 @@ async function waitForServer(
         );
         resolve(undefined);
       };
+      client.on("close", onClose);
       client.on("open", () => {
         client.off("close", onClose);
         signal.addEventListener("abort", () => {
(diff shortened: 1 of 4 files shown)


typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -0,0 +1,221 @@
...
+      });
+      client.on("status", (event) => {
+        log(event);
+      });
+      client.on("error", (err) => {
+        log("server error: %o", err);
+      });
+
+      client.on("advertise", (newChannels) => {
+        void Promise.all(
```
achim-k, line 126: is the `void` necessary here? What purpose does it serve?

achim-k (approved):

typescript/ws-protocol-examples/src/examples/mcap-record.ts
jtbandes: mostly to keep the linter happy, since we are mixing async and non-async functions here

PR #444, opened 2023-05-17
Python: Disable compression by default, allow specifying server kwargs

### LINKED ISSUE #443: Websocket connection performance - Make compression a parameter for `FoxgloveServer`
The Python Websocket API enables compression by default. When sending large data, this leads to a significant performance decrease and CPU load increase at the same time.
It would be great to have it as a parameter when constructing `FoxgloveServer`, in order to be able to deactivate it.

It might make sense (probably) to be able to manually set this for all language bindings.

achim-k:
### Public-Facing Changes

- Python: Disable compression by default, allow specifying server kwargs

### Description
Allows specifying [websockets.server.serve(...) ](https://websockets.readthedocs.io/en/stable/reference/asyncio/server.html#websockets.server.serve)keyword arguments and disables compression (`compression=None`) by default

Fixes #443

### DIFF
--- a/python/src/foxglove_websocket/server/__init__.py
+++ b/python/src/foxglove_websocket/server/__init__.py
@@ -163,7 +163,7 @@ class FoxgloveServer:
         metadata: Optional[Mapping[str, str]] = None,
         session_id: Optional[str] = None,
         logger: logging.Logger = _get_default_logger(),
-        server_kwargs: Dict[str, Any] = { "compression": None },
+        server_kwargs: Dict[str, Any] = {"compression": None},
     ):
         self.host = host
         self.port = port
@@ -224,7 +224,7 @@ class FoxgloveServer:
                 self.host,
                 self.port,
                 subprotocols=[Subprotocol("foxglove.websocket.v1")],
-                **self._server_kwargs
+                **self._server_kwargs,
             )
             self._opened.set_result(server)
         except asyncio.CancelledError:
--- a/python/src/foxglove_websocket/server/__init__.py
+++ b/python/src/foxglove_websocket/server/__init__.py
@@ -163,6 +163,7 @@ class FoxgloveServer:
         metadata: Optional[Mapping[str, str]] = None,
         session_id: Optional[str] = None,
         logger: logging.Logger = _get_default_logger(),
+        server_kwargs: Dict[str, Any] = { "compression": None },
     ):
         self.host = host
         self.port = port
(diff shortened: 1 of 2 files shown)


python/src/foxglove_websocket/server/__init__.py
```diff
@@ -163,6 +163,7 @@ def __init__(
         metadata: Optional[Mapping[str, str]] = None,
         session_id: Optional[str] = None,
         logger: logging.Logger = _get_default_logger(),
+        server_kwargs: Dict[str, Any] = {"compression": None},
```
jtbandes, line 166: Is there something we can do to pass through the kwarg types from the serve function so autocomplete and type checking work properly?

jtbandes (approved):

python/src/foxglove_websocket/server/__init__.py
KerstinKeller: Apparently you can inspect the signature of a function/object (e.g. serve) such as
```
>>> from websockets.server import serve
>>> import inspect
>>> inspect.signature(server)
<Signature (ws_handler: 'Union[Callable[[WebSocketServerProtocol], Awaitable[Any]], Callable[[WebSocketServerProtocol, str], Awaitable[Any]]]', host: 'Optional[Union[str, Sequence[str]]]' = None, port: 'Optional[int]' = None, *, create_protocol: 'Optional[Callable[[Any], WebSocketServerProtocol]]' = None, logger: 'Optional[LoggerLike]' = None, compression: 'Optional[str]' = 'deflate', origins: 'Optional[Sequence[Optional[Origin]]]' = None, extensions: 'Optional[Sequence[ServerExtensionFactory]]' = None, subprotocols: 'Optional[Sequence[Subprotocol]]' = None, extra_headers: 'Optional[HeadersLikeOrCallable]' = None, process_request: 'Optional[Callable[[str, Headers], Awaitable[Optional[HTTPResponse]]]]' = None, select_subprotocol: 'Optional[Callable[[Sequence[Subprotocol], Sequence[Subprotocol]], Subprotocol]]' = None, ping_interval: 'Optional[float]' = 20, ping_timeout: 'Optional[float]' = 20, close_timeout: 'Optional[float]' = None, max_size: 'Optional[int]' = 1048576, max_queue: 'Optional[int]' = 32, read_limit: 'int' = 65536, write_limit: 'int' = 65536, **kwargs: 'Any') -> 'None'>
```
which you can very problably use somethow as a type for `server_kwarg`.

However, some of the args are set manually (`self.host`, `self.port`, ...) and not part of these `server_kwargs`.

PR #462, opened 2023-06-05
Extend spec to allow fetching of assets

achim-k:
### Public-Facing Changes

- Extend spec to allow fetching of assets

### Description
Allows clients to request assets (files) from the server.

Example use case: Client requests mesh files which are referenced by a URDF

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -505,15 +505,15 @@ Fetch an asset from the server. Only supported if the server previously declared
 #### Fields
 
 - `op`: string `"fetchAsset"`
-- `assetId`: string, unique identifier to locate the asset
-- `requestId`: number, unique 32-bit unsigned integer which is to be included in the response
+- `assetURI`: string, uniform resource identifier to locate a single asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the [response](#fetch-asset-response)
 
 #### Example
 
 ```json
 {
   "op": "fetchAsset",
-  "assetId": "file:///foo/bar.txt",
+  "assetURI": "package://foo/robot.urdf",
   "requestId": 123
 }
 ```
@@ -566,14 +566,16 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 - Response to a [fetch asset](#fetch-asset) request
 - Only supported if the server previously declared the `assets` [capability](#server-info).
 
-| Bytes           | Type             | Description                                                                                                             |
-| --------------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------- |
-| 1               | opcode           | 0x04                                                                                                                    |
-| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                       |
-| 1               | uint8            | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                 |
-| 4 + N           | uint32 + char[N] | error message, empty if `status == 0`                                                                                   |
-| 4 + M           | uint32 + char[M] | asset [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), empty if `status != 0` |
-| remaining bytes | uint8[]          | asset data (file contents), empty if `status != 0`                                                                      |
+| Bytes                  | Type    | Description                                                                                                             |
+| ---------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------- |
+| 1                      | opcode  | 0x04                                                                                                                    |
+| 4                      | uint32  | request id, as given in the corresponding [request](#fetch-asset)                                                       |
+| 1                      | uint8   | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                 |
+| 4                      | uint32  | error message length                                                                                                    |
+| _error message length_ | char[]  | error message, empty if `status == 0`                                                                                   |
+| 4                      | uint32  | media type length                                                                                                       |
+| _media type length_    | char[]  | asset [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), empty if `status != 0` |
+| remaining bytes        | uint8[] | asset data (file contents), empty if `status != 0`                                                                      |
 
 ### Client Message Data
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -566,13 +566,13 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 - Response to a [fetch asset](#fetch-asset) request
 - Only supported if the server previously declared the `assets` [capability](#server-info).
 
-| Bytes           | Type             | Description                                                                                                                                                   |
-| --------------- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| 1               | opcode           | 0x04                                                                                                                                                          |
-| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                                                             |
-| 1               | uint8            | status, value is `1` if the asset was fetched succesfully, all other values are error codes                                                                   |
-| 4 + N           | uint32 + char[N] | [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) of the asset (if `status == 1`) or error message (if `status != 1`) |
-| remaining bytes | uint8[]          | asset data (file contents, empty if `status != 1`)                                                                                                            |
+| Bytes           | Type             | Description                                                                                                                                              |
+| --------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1               | opcode           | 0x04                                                                                                                                                     |
+| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                                                        |
+| 1               | uint8            | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                                                  |
+| 4 + N           | uint32 + char[N] | asset [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), if `status == 0` <br /> error message, if `status == 1` |
+| remaining bytes | uint8[]          | asset data (file contents), empty if `status != 0`                                                                                                       |
 
 ### Client Message Data
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -573,8 +573,6 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
 | 1                      | uint8   | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                 |
 | 4                      | uint32  | error message length                                                                                                    |
 | _error message length_ | char[]  | error message, empty if `status == 0`                                                                                   |
-| 4                      | uint32  | media type length                                                                                                       |
-| _media type length_    | char[]  | asset [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), empty if `status != 0` |
 | remaining bytes        | uint8[] | asset data (file contents), empty if `status != 0`                                                                      |
 
 ### Client Message Data
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -505,7 +505,7 @@ Fetch an asset from the server. Only supported if the server previously declared
 #### Fields
 
 - `op`: string `"fetchAsset"`
-- `assetURI`: string, uniform resource identifier to locate a single asset
+- `assetUri`: string, uniform resource identifier to locate a single asset
 - `requestId`: number, unique 32-bit unsigned integer which is to be included in the [response](#fetch-asset-response)
 
 #### Example
@@ -513,7 +513,7 @@ Fetch an asset from the server. Only supported if the server previously declared
 ```json
 {
   "op": "fetchAsset",
-  "assetURI": "package://foo/robot.urdf",
+  "assetUri": "package://foo/robot.urdf",
   "requestId": 123
 }
 ```
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -34,6 +34,7 @@
 - [Unadvertise Services](#unadvertise-services) (json)
 - [Service Call Response](#service-call-response) (binary)
 - [Connection Graph Update](#connection-graph-update) (json)
+- [Asset](#asset) (binary)
 
 ### Sent by client
 
@@ -49,6 +50,7 @@
 - [Service Call Request](#service-call-request) (binary)
 - [Subscribe connection graph updates](#subscribe-connection-graph-updates) (json)
 - [Unubscribe connection graph updates](#unsubscribe-connection-graph-updates) (json)
+- [Fetch asset](#fetch-asset) (json)
 
 ## JSON messages
 
@@ -69,6 +71,7 @@ Each JSON message must be an object containing a field called `op` which identif
   - `time`: The server may publish binary [time](#time) messages
   - `services`: Allow clients to call services
   - `connectionGraph`: Allow clients to subscribe to updates to the connection graph
+  - `assets`: Allow clients to fetch assets
 - `supportedEncodings`: array of strings | informing the client about which encodings may be used for client-side publishing or service call requests/responses. Only set if client publishing or services are supported.
 - `metadata`: optional map of key-value pairs
 - `sessionId`: optional string. Allows the client to understand if the connection is a re-connection or if it is connecting to a new server instance. This can for example be a timestamp or a UUID.
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
 }
 ```
 
+### Fetch asset
+
+Fetch an asset from the server. Only supported if the server previously declared that it has the `assets` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"fetchAsset"`
+- `uri`: string, unique identifier to locate the asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the response
+
+#### Example
+
+```json
+{
+  "op": "fetchAsset",
+  "uri": "file:///foo/bar.txt",
+  "requestId": 123
+}
+```
+
 ## Binary messages
 
 All binary messages must start with a 1-byte opcode identifying the type of message. The interpretation of the remaining bytes depends on the opcode.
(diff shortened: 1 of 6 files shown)


docs/spec.md
```diff
@@ -538,6 +561,20 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+| Bytes               | Type    | Description                                                                                            |
+| ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| 1                   | opcode  | 0x04                                                                                                   |
+| 4                   | uint32  | request id, as given in the corresponding [request](#fetch-asset)                                      |
+| 8                   | uint64  | Last modified time of the asset (nanoseconds)                                                          |
+| 4                   | uint32  | media type length                                                                                      |
+| _media type length_ | char[]  | [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) of the asset |
+| remaining bytes     | uint8[] | asset data (file contents)                                                                             |
+
```
achim-k, line 577: In case the asset can't be fetched (e.g. does not exist), should we sent a special response? Right now that's not the case, and the client would have to determine itself (e.g. through a timeout) if the asset failed to fetch

docs/spec.md
```diff
@@ -538,6 +561,20 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+### Asset
+
+- Response to a [fetch asset](#fetch-asset) request if the asset was found
+- Only supported if the server previously declared the `assets` [capability](#server-info).
+
+| Bytes               | Type    | Description                                                                                            |
+| ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| 1                   | opcode  | 0x04                                                                                                   |
+| 4                   | uint32  | request id, as given in the corresponding [request](#fetch-asset)                                      |
+| 8                   | uint64  | Last modified time of the asset (nanoseconds)                                                          |
```
achim-k, line 573: Not sure if the last modified time is very relevant for the client. Maybe we can omit it

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
 ```
 
+### Fetch asset
+
+Fetch an asset from the server. Only supported if the server previously declared that it has the `assets` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"fetchAsset"`
+- `uri`: string, unique identifier to locate the asset
```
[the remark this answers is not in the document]
achim-k: No, it does not have to be an identifier following the URI spec, it can be any string that identifies an asset uniquely. Not sure what a better field name would be, maybe just `id` ?

docs/spec.md
```diff
@@ -538,6 +561,20 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+| Bytes               | Type    | Description                                                                                            |
+| ------------------- | ------- | ------------------------------------------------------------------------------------------------------ |
+| 1                   | opcode  | 0x04                                                                                                   |
+| 4                   | uint32  | request id, as given in the corresponding [request](#fetch-asset)                                      |
+| 8                   | uint64  | Last modified time of the asset (nanoseconds)                                                          |
+| 4                   | uint32  | media type length                                                                                      |
+| _media type length_ | char[]  | [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) of the asset |
+| remaining bytes     | uint8[] | asset data (file contents)                                                                             |
```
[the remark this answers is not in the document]
achim-k: > How do you think it will work if the asset is very large? Will the websocket connection be able to handle it?

If the asset response message would exceed the max. websocket frame size, the message would be split up into multiple frames and send like that to the client. So there is theoretically no limit in file size

> I wonder if it would make sense to have an option where the server could return a url that the client can fetch() to load the asset in a separate request?

We could maybe use the `status` field for that:

```ts
enum FetchAssetStatus {
  SUCCESS_ASSET_DATA = 1,  // Request contains binary asset data
  SUCCESS_ASSET_URL = 2,    // Request contains asset URL
  // Other values: Error
}
```

docs/spec.md
achim-k: can't really come up with a use case, will remove it

docs/spec.md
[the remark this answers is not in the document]
jtbandes: What would the client do with that info? At the end of the day the server is going to be the one who decides whether the request is valid or not.

docs/spec.md
```diff
@@ -538,6 +561,19 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+### Asset
+
+- Response to a [fetch asset](#fetch-asset) request
+- Only supported if the server previously declared the `assets` [capability](#server-info).
+
+| Bytes           | Type             | Description                                                                                                                                                   |
+| --------------- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1               | opcode           | 0x04                                                                                                                                                          |
+| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                                                             |
+| 1               | uint8            | status, value is `1` if the asset was fetched succesfully, all other values are error codes                                                                   |
```
jtbandes, line 573: should we make this a proper enum and say that `0` is success, `1` is error, and >1 are reserved values? (or 0=error and 1=success or something?) Then we could expand to different types of responses later without separate op codes.

docs/spec.md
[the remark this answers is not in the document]
achim-k: The client could use this info to determine what a valid url would be. I was thinking e.g. about [this](https://github.com/foxglove/studio/blob/28ab637201c23163980f0bdd2145a27ddc532e3a/packages/studio-base/src/panels/ThreeDeeRender/renderables/Urdfs.ts#L823-L832)

docs/spec.md
achim-k: OK I will do that. 

> (or 0=error and 1=success or something?)

I prefer this, since an uninitialized value would cause an error

typescript/ws-protocol/src/FoxgloveServer.ts
```diff
@@ -599,4 +609,39 @@ export default class FoxgloveServer {
...
+    view.setUint8(offset, response.status);
+    offset += 1;
+    view.setUint32(offset, mediaTypeOrErrMsg.length, true);
+    offset += 4;
+    payload.set(mediaTypeOrErrMsg, offset);
+    offset += mediaTypeOrErrMsg.length;
+
+    if (isSuccess) {
+      payload.set(
+        new Uint8Array(response.data.buffer, response.data.byteOffset, response.data.byteLength),
```
jtbandes, line 640: why not just `response.data`?

jtbandes (approved):
Looks good to me, I would be interested to get @defunctzombie's take on it as well

typescript/ws-protocol/src/FoxgloveServer.ts
achim-k: `response.data` is a `DataView`, so it gives

```
Argument of type 'DataView' is not assignable to parameter of type 'ArrayLike<number>'
```

I guess that I should have used `Uint8Array` in the type definition, but I'm using `DataView` also for other types, so at least it's consistent.

typescript/ws-protocol/src/types.ts
```diff
@@ -2,6 +2,7 @@ export enum BinaryOpcode {
   MESSAGE_DATA = 1,
   TIME = 2,
   SERVICE_CALL_RESPONSE = 3,
+  ASSET = 4,
```
achim-k, line 5:
```suggestion
  FETCH_ASSET_RESPONSE = 4,
```

Maybe more suited, now that the response can also be a error response

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
+- `assetId`: string, unique identifier to locate the asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the response
+
+#### Example
+
+```json
+{
+  "op": "fetchAsset",
+  "assetId": "file:///foo/bar.txt",
+  "requestId": 123
```
defunctzombie, line 517: I don't see us using `requestId` for other opcodes. Why here?

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
+- `op`: string `"fetchAsset"`
+- `assetId`: string, unique identifier to locate the asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the response
+
+#### Example
+
+```json
+{
+  "op": "fetchAsset",
+  "assetId": "file:///foo/bar.txt",
```
defunctzombie, line 516: I find it strange that in the studio PR this is the asset name yet here you are calling it the _id_.

docs/spec.md
```diff
@@ -538,6 +561,19 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+| Bytes           | Type             | Description                                                                                                                                              |
+| --------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1               | opcode           | 0x04                                                                                                                                                     |
+| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                                                        |
+| 1               | uint8            | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                                                  |
+| 4 + N           | uint32 + char[N] | asset [media type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), if `status == 0` <br /> error message, if `status == 1` |
```
defunctzombie, line 574: I find this a bit weird that this field is overloaded with the error message and the media type. Is there a problem with using a separate error field that is empty if there is no error? The length prefix overhead?

Another option would be two opcodes (one for assets success and another for error) not sure if this was discussed.

defunctzombie:
Left a few comments - I'd like to work through some different scenarios that we'd experience in studio i.e. local data w/layout, then using layout for live, and also data platform access so we are clear on what the various asset names can be in those cases. I don't feel like I have a good intuition that we've covered a cohesive story around these workflows. It might turn out that there is no single story so we'll see.

docs/spec.md
achim-k: > Is there a problem with using a separate error field that is empty if there is no error? The length prefix overhead?

Changed it to use a separate field.

docs/spec.md
defunctzombie: What's the purpose of the requestId for this request? can I omit it?

docs/spec.md
```diff
@@ -538,6 +561,20 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+- Response to a [fetch asset](#fetch-asset) request
+- Only supported if the server previously declared the `assets` [capability](#server-info).
+
+| Bytes           | Type             | Description                                                                                                             |
+| --------------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------- |
+| 1               | opcode           | 0x04                                                                                                                    |
+| 4               | uint32           | request id, as given in the corresponding [request](#fetch-asset)                                                       |
+| 1               | uint8            | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                 |
+| 4 + N           | uint32 + char[N] | error message, empty if `status == 0`                                                                                   |
```
defunctzombie, line 574: The way these response fields are documented does not match how we've documented this kind of length prefixed response for other kinds of responses:

https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#service-call-response

Consistency is important for spec sanity so we should pick the style we want. I'd say this PR should follow the status-quo and a separate PR can propose altering the status-quo.

defunctzombie (commented):
I still find the use of `assetId` in the request somewhat "awkward" if the intent is to load an asset by URI/URL.

I'm inclined to say that imposing the requirement that you fetch assets by URI sane given that this protocol is to power interfacing with Studio and this seems like a sane way to describe references to assets in Studio.

I do think there is a workflow where you lookup by "id" but to me that feels like it would be an "opaque" ID rather than a lookup by human readable?

---
Does fetch asset return a single asset? What happens if there are multiple for a given "uri"? Is one picked randomly? Error?

Maybe for websocket/live connections it is more obvious that there should be one asset for a uri but for mcap we allow the same name to appear for multiple attachments. If you imagine that the name would be the uri (i.e. `package://foo.urdf`) then there's an open question on what the Studio APIs for fetching an asset would be like. Maybe that's a problem for studio but worth considering whether that flexibility needs to find its way to the ws-protocol.

All of that is to say maybe the opcodes could be specific like `fetchAssetById` `fetchAssetByUri`. Another common term used from ORM lands is `findOneBy` vs `find`. Not necessarily saying we should adopt these terms here but instead offering them as other kinds of prior art for "finding stuff".

achim-k:
> I still find the use of `assetId` in the request somewhat "awkward" if the intent is to load an asset by URI/URL.
> 
> I'm inclined to say that imposing the requirement that you fetch assets by URI sane given that this protocol is to power interfacing with Studio and this seems like a sane way to describe references to assets in Studio.

Yes, I would propose to do exactly that. 

> Does fetch asset return a single asset? What happens if there are multiple for a given "uri"? Is one picked randomly? Error?

When fetching by an identifier (such as URI), only a single asset can be returned (or error if the asset is not found). If we decide to later allow fetching assets by name, this may be different.

> Maybe for websocket/live connections it is more obvious that there should be one asset for a uri but for mcap we allow the same name to appear for multiple attachments. If you imagine that the name would be the uri (i.e. `package://foo.urdf`) then there's an open question on what the Studio APIs for fetching an asset would be like. Maybe that's a problem for studio but worth considering whether that flexibility needs to find its way to the ws-protocol.
> 
> All of that is to say maybe the opcodes could be specific like `fetchAssetById` `fetchAssetByUri`. Another common term used from ORM lands is `findOneBy` vs `find`. Not necessarily saying we should adopt these terms here but instead offering them as other kinds of prior art for "finding stuff".

I don't think we should be influenced here too much how assets (attachments) are stored & retrieved in MCAP. At least I don't see why this should prevent us from enforcing that the `fetchAsset` (by URI) operation only returns a single asset. If we realize that this operation is not sufficient, we can easily add another `fetchAssetsByName` operation as you have noted.

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
 
+### Fetch asset
+
+Fetch an asset from the server. Only supported if the server previously declared that it has the `assets` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"fetchAsset"`
+- `assetURI`: string, uniform resource identifier to locate a single asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the [response](#fetch-asset-response)
```
jtbandes, line 509: nit: inconsistent capitalization between `URI` & `Id`

jtbandes (approved):

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
 ```
 
+### Fetch asset
+
+Fetch an asset from the server. Only supported if the server previously declared that it has the `assets` [capability](#server-info).
+
+#### Fields
+
+- `op`: string `"fetchAsset"`
+- `assetURI`: string, uniform resource identifier to locate a single asset
```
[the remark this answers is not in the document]
jhurliman: > However, I think it could be worth calling out that different servers may interpret URIs in different ways

That's exactly how URIs work, and one of the distinguishing features from the more specific subset of URIs known as URLs. `tel:+1-816-555-1212` is a valid URI but it doesn't say anything about how to "fetch" that resource, or if "fetch" is even a valid verb for that resource in the context of a given server.

I support the requirement that assets are identified by a URI.

docs/spec.md
```diff
@@ -495,6 +498,26 @@ Unsubscribe from [connection graph updates](#connection-graph-update). Only supp
...
+- `op`: string `"fetchAsset"`
+- `assetUri`: string, uniform resource identifier to locate a single asset
+- `requestId`: number, unique 32-bit unsigned integer which is to be included in the [response](#fetch-asset-response)
+
+#### Example
+
+```json
+{
+  "op": "fetchAsset",
+  "assetUri": "package://foo/robot.urdf",
```
jtbandes, line 516: Maybe this should just be called `uri`? The op is already referring to an asset.

typescript/ws-protocol/src/FoxgloveServer.test.ts
```diff
@@ -480,4 +487,83 @@ describe("FoxgloveServer", () => {
     }
     close();
   });
+
+  it.each([
```
jtbandes, line 491: nit: I think you can do something like `it.each<FetchAssetResponse>(...)` and remove `as`

typescript/ws-protocol/src/types.ts
```diff
@@ -188,6 +200,19 @@ export type Time = {
...
+  op: BinaryOpcode.FETCH_ASSET_RESPONSE;
+  requestId: number;
+  status: FetchAssetStatus.SUCCESS;
+  data: DataView;
+};
+export type FetchAssetErrorResponse = {
+  op: BinaryOpcode.FETCH_ASSET_RESPONSE;
+  requestId: number;
+  status: FetchAssetStatus.ERROR;
+  errorMsg: string;
```
jtbandes, line 213: maybe name this `error: string`? I find the "msg" abbreviation a bit too informal for this API.

jtbandes (approved):
Left some comments regarding name tweaks.

docs/spec.md
```diff
@@ -538,6 +561,20 @@ All integer types explicitly specified (uint32, uint64, etc.) in this section ar
...
+
+- Response to a [fetch asset](#fetch-asset) request
+- Only supported if the server previously declared the `assets` [capability](#server-info).
+
+| Bytes                  | Type    | Description                                                                                                             |
+| ---------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------- |
+| 1                      | opcode  | 0x04                                                                                                                    |
+| 4                      | uint32  | request id, as given in the corresponding [request](#fetch-asset)                                                       |
+| 1                      | uint8   | status enum, `0` for success and `1` for error. Values `>1` are reserved for future use                                 |
+| 4                      | uint32  | error message length                                                                                                    |
```
defunctzombie, line 574: not sure if worth it - but 64k ought to be enough for any error message 🤣

defunctzombie (approved):

docs/spec.md
defunctzombie: I suppose uint32 is what we use as the length prefix for strings consistently so this seems 👍

PR #463, opened 2023-06-09
Add optional `schema`  & `schemaEncoding` fields to client advertise operation

achim-k:
### Public-Facing Changes

Add optional `schema`  & `schemaEncoding` fields to client advertise operation

### Description
See https://github.com/orgs/foxglove/discussions/728

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -326,8 +326,10 @@ Informs the client about updates to the connection graph. This is only sent to c
 - `channels`: array of:
   - `id`: number chosen by the client. The client may reuse ids that have previously been unadvertised.
   - `topic`: string
-  - `encoding`: string, one of the encodings [supported by the server](#server-info)
+  - `encoding`: string, one of the message encodings [supported by the server](#server-info)
   - `schemaName`: string
+  - `schema`: string | undefined, optional [schema definition](#advertise).
+  - `schemaEncoding`: string | undefined, optional [schema encoding](#advertise).
 
 #### Example
 
--- a/typescript/ws-protocol/src/FoxgloveClient.ts
+++ b/typescript/ws-protocol/src/FoxgloveClient.ts
@@ -164,9 +164,15 @@ export default class FoxgloveClient {
     this.send({ op: "unsubscribe", subscriptionIds: [subscriptionId] });
   }
 
-  advertise(topic: string, encoding: string, schemaName: string): ClientChannelId {
+  advertise(
+    topic: string,
+    encoding: string,
+    schemaName: string,
+    schema?: string,
+    schemaEncoding?: string,
+  ): ClientChannelId {
     const id = ++this.nextAdvertisementId;
-    const channels: ClientChannel[] = [{ id, topic, encoding, schemaName }];
+    const channels: ClientChannel[] = [{ id, topic, encoding, schemaName, schema, schemaEncoding }];
     this.send({ op: "advertise", channels });
     return id;
   }
--- a/typescript/ws-protocol/src/types.ts
+++ b/typescript/ws-protocol/src/types.ts
@@ -59,6 +59,8 @@ export type ClientChannel = {
   topic: string;
   encoding: string;
   schemaName: string;
+  schema?: string;
+  schemaEncoding?: string;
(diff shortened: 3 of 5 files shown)


typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -164,9 +164,15 @@ export default class FoxgloveClient {
     this.send({ op: "unsubscribe", subscriptionIds: [subscriptionId] });
   }
 
-  advertise(topic: string, encoding: string, schemaName: string): ClientChannelId {
+  advertise(
+    topic: string,
+    encoding: string,
+    schemaName: string,
+    schema?: string,
+    schemaEncoding?: string,
```
jtbandes, line 172: Feels like this is too many string params in a row, should become an object

docs/spec.md
```diff
@@ -326,8 +326,10 @@ Informs the client about updates to the connection graph. This is only sent to c
 - `channels`: array of:
   - `id`: number chosen by the client. The client may reuse ids that have previously been unadvertised.
   - `topic`: string
-  - `encoding`: string, one of the encodings [supported by the server](#server-info)
+  - `encoding`: string, one of the message encodings [supported by the server](#server-info)
   - `schemaName`: string
+  - `schema`: string | undefined, optional [schema definition](#advertise).
+  - `schemaEncoding`: string | undefined, optional [schema encoding](#advertise).
```
jtbandes, line 332: Should we say schemaEncoding is required if schema is set?

jtbandes (approved):

typescript/ws-protocol/src/FoxgloveClient.ts
achim-k: I was thinking the same, but didn't want to break API. Will deprecate this method and add one with an object

docs/spec.md
achim-k: :+1: 
Is there a way to enforce this with typescript?

jtbandes: You might have to do something like `{ ...; schema: string; schemaEncoding: string; } | { ...; schema?: undefined; schemaEncoding?: undefined }`

typescript/ws-protocol/src/FoxgloveClient.ts
jtbandes: We could also bump to v0.6 and break the API :)

docs/spec.md
achim-k: I tried this, but this doesn't work well together with `Omit`: https://www.typescriptlang.org/play?#code/C4TwDgpgBAwgNgSwgO2DAFgQ2cicoC8UA3gFBRQIAmAXFMgK4C2ARhAE4Dc5UwA9mAQBjOgGdg7BMgDm3CiiF8qU6WIkq5UUUPQQmmAHKYmENZJncAvlABkUABTEtOvZjMbnu-QFFki5TLuMlDWAD4knq50DMhUEABmUhBUnJE+fkoq0bEJSVQhAJTcpKTxMULACHzIUJhUAG4claIQ9kKIKGhYOHhQdADyTAjAADzwSKgY2LhwADRQAETUCwB8BSQ8FFCKyOKU+UQAjJpbO3u4AO5TPfh0453XM4QRAHRv7RNd03jz1CGpmwoZz4cAgLzgfGk9kujzwRVIliAA

typescript/ws-protocol/src/types.ts
```diff
@@ -53,13 +53,19 @@ export type Unsubscribe = {
   op: "unsubscribe";
   subscriptionIds: SubscriptionId[];
 };
-
-export type ClientChannel = {
+type None<T> = { [K in keyof T]?: never };
+type BothOrNone<T1, T2> = (None<T1> & None<T2>) | (T1 & T2);
```
jtbandes, line 57: I might go with `type ClientChannelSchemaInfo = { schema: string, schemaEncoding: string } | { schema?: undefined, schemaEncoding?: undefined }` instead of the generic stuff for simplicity.

jtbandes (approved):

PR #469, opened 2023-06-15
Add C++ FlatBuffer example server

jtbandes:


achim-k (approved):
Nice!

cpp/examples/src/example_server_flatbuffers.cpp
```diff
@@ -0,0 +1,141 @@
...
+      builder, foxglove::CreateVector3(builder, 2, 0, 0),
+      createQuaternionFromAxisAngle(builder, 0, 0, 1, double(now) / 1e9 * 0.5));
+    auto size = foxglove::CreateVector3(builder, 1, 1, 1);
+    auto color = foxglove::CreateColor(builder, 0.6, 0.2, 1, 1);
+    auto cubeBuilder = foxglove::CubePrimitiveBuilder(builder);
+    cubeBuilder.add_pose(pose);
+    cubeBuilder.add_size(size);
+    cubeBuilder.add_color(color);
+    const auto cube = cubeBuilder.Finish();
+    builder.Finish(cube);
```
jameskuszmaul-brt, line 115: This call is unneeded (and may be harmful?). You should only call `builder.Finish` at the end.

Your problems were just that you had been building more than one flatbuffer table at once, when you are suppose to have called Finish (or called Create*, which calls Finish under the hood) before you start creating the new table.

snosenzo (approved):
echoing james' point too you should fix the pre-emptive `.Finish`es on the builder

PR #498, opened 2023-07-17
Add dependabot group for eslint packages

achim-k:
### Public-Facing Changes

None

### Description
Adds a dependabot group to group eslint package updates together in one PR

From https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups:

> By default, Dependabot raises a single pull request for each dependency that needs to be updated to a newer version. You can use groups to create sets of dependencies (per package manager), so that Dependabot opens a single pull request to update multiple dependencies at the same time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,13 @@ updates:
       interval: weekly
     labels: [] # disable default labels
 
+    # Define groups of dependencies to be updated together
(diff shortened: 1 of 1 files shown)


jtbandes (approved):
👍

achim-k:
@jtbandes should `@foxglove/eslint-plugin` be included into the group as well?

PR #519, opened 2023-08-08
Add parameter type `float64` & `float64_array` for floating point values

achim-k:
### Public-Facing Changes
Add parameter type `float64` & `float64_array` for floating point values

### Description
Some languages, such as JavaScript, do not differentiate between integers and floating point numbers. The `float64` and `float64_array` types were added to avoid ambiguity by informing the client (or the server) that the value is a floating point number, or an array of it, respectively.

Example: When sending the following JSON via JS to the server, the server will receive the value `1` and deduce that the parameter is of type Integer.

```json
{
  "op": "setParameters",
  "parameters": [
    { "name": "/float_param", "value": 1.0 }
  ]
}
```

By adding `"type": "float64"`, the client can let the server know that the value should be treated as a 64-bit floating point number.

Related to FG-4486

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -174,7 +174,7 @@ Informs the client about parameters. Only supported if the server declares the `
 - `parameters`: array of:
   - `name`: string, name of the parameter
   - `value`: ParameterValue, where ParameterValue is of type number | boolean | string | ParameterValue[] | { [key: string]: ParameterValue } | undefined.
-  - `type`: "byte_array" | undefined. If the type is `byte_array`, `value` shall be interpreted as base64 encoded string.
+  - `type`: "byte_array" | "float64" | "float64_array" | undefined. If the type is `byte_array`, `value` shall be interpreted as base64 encoded string. If the type is `float64` or `float64_array`, `value` shall be a 64-bit floating-point number or integer, or an array of them, respectively.
 - `id`: string | undefined. Only set when the [getParameters](#get-parameters) or [setParameters](#set-parameters) request's `id` field was set
 
 #### Example
@@ -188,6 +188,7 @@ Informs the client about parameters. Only supported if the server declares the `
     { "name": "/string_param", "value": "foo" },
     { "name": "/node/nested_ints_param", "value": [1, 2, 3] }
     { "name": "/byte_array_param", "value": "QUJDRA==", "type": "byte_array" },
+    { "name": "/float_param_int", "value": 3, "type": "float64" },
   ],
   "id": "request-123"
 }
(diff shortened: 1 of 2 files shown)


docs/spec.md
```diff
@@ -414,7 +415,8 @@ Set one or more parameters. Only supported if the server previously declared tha
   "parameters": [
     { "name": "/int_param", "value": 3 },
     { "name": "/float_param", "value": 4.1 },
-    { "name": "/byte_array_param", "value": "QUJDRA==", "type": "byte_array" }
+    { "name": "/byte_array_param", "value": "QUJDRA==", "type": "byte_array" },
+    { "name": "/float_array_param", "value": [1.1, 2, 3.3], "type": "float64_array" }
```
snosenzo, line 419: Should both `float64` and `float64_array` lines be in  both examples?
Unclear to me why not have them in both fi the byte_array is in both

snosenzo (approved):

docs/spec.md
achim-k: agreed, will do

snosenzo (approved):

defunctzombie:
I didn't catch this sooner but I don't understand the following spec:

> If the type is `float64` or `float64_array`, `value` shall be a 64-bit floating-point number or integer,

What is "or integer" about? If it is a floating point value it is not an integer.

What about `float32`?

> or an array of them respecitvely

What does it mean to have an array of them? What would the value be?

---
As for the original issue that was reported - why does the receiver not figure out that the parameter type is a double and treat the incoming number as a double?

achim-k:
> I didn't catch this sooner but I don't understand the following spec:
> 
> > If the type is `float64` or `float64_array`, `value` shall be a 64-bit floating-point number or integer,
> 
> What is "or integer" about? If it is a floating point value it is not an integer.
> 
> What about `float32`?

In the end everything that is convertible to a 64-bit floating point number is allowed. So yes, also a `float32`. 

> 
> > or an array of them respecitvely
> 
> What does it mean to have an array of them? What would the value be?

An array of floating point numbers, integers or a mix of it:

```js
[1.1, 2, 3.3, 4, 5.5]
```

> As for the original issue that was reported - why does the receiver not figure out that the parameter type is a double and treat the incoming number as a double?

This would work for existing parameters, but not if you set a new parameter.

defunctzombie:
Thanks for clarifying. Based on what you described I think the language in the spec:

> If the type is `float64` or `float64_array`, `value` shall be a 64-bit floating-point number or integer, or an array of them, respectively.

Could be clearer. I now understand that when using the "float64" type you allow either a decimal number or a number without a decimal (i.e. integer). I propose something like:

> If the type is `float64`, `value` must be a valid decimal or integer value that can be represented by a float64. If the type is `float64_array`, `value` must be an array of valid decimal or integer values.

I personally found the current language in the spec (in this PR) that has the two types (float64 and float64_array) and _then_ says value can be an array confusing. It made me think that for float64 the value could be an array.

PR #527, opened 2023-08-21
Fix CI check for using current ws-protocol package not working

achim-k:
### Public-Facing Changes
None

### Description
Fixes the `Ensure examples use current version of ws-protocol package` CI check not working (as no `yarn.lock` file existed)


achim-k:
You are absolutely right, this PR doesn't make any sense

PR #538, opened 2023-09-11
Upgrade typescript-eslint to v6

achim-k:
### Public-Facing Changes

None

### Description
Basically does the same as https://github.com/foxglove/studio/pull/6724
> Upgrades typescript-eslint to v6, and uses the new @foxglove/eslint-plugin that's compatible with it.

### DIFF
--- a/typescript/ws-protocol/src/parse.ts
+++ b/typescript/ws-protocol/src/parse.ts
@@ -53,11 +53,9 @@ export function parseServerMessage(buffer: ArrayBuffer): ServerMessage {
 
       if (status === FetchAssetStatus.ERROR) {
         return { op, requestId, status, error };
-      } else if (status === FetchAssetStatus.SUCCESS) {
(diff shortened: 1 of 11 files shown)


typescript/ws-protocol/src/parse.ts
```diff
@@ -53,11 +53,9 @@ export function parseServerMessage(buffer: ArrayBuffer): ServerMessage {
 
       if (status === FetchAssetStatus.ERROR) {
         return { op, requestId, status, error };
-      } else if (status === FetchAssetStatus.SUCCESS) {
+      } else {
         const data = new DataView(buffer, offset, buffer.byteLength - offset);
         return { op, requestId, status, data };
-      } else {
```
jtbandes, line 59: Might want to keep this just to be safe

jtbandes (approved):
Thanks!

PR #539, opened 2023-09-12
Add pure CMake support

souryavarenya:
### Public-Facing Changes

- Adds CMake support

### Description

We internally use CMake-based tooling to build packages from source and install. This PR improves the existing `CMakeLists.txt` to install CMake config.

### DIFF
--- a/cpp/foxglove-websocket/CMakeLists.txt
+++ b/cpp/foxglove-websocket/CMakeLists.txt
@@ -1,15 +1,85 @@
+# General Info
 cmake_minimum_required(VERSION 3.15)
-project(FoxgloveWebSocket CXX)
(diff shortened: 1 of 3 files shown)


achim-k:
Does https://github.com/foxglove/ws-protocol/blob/main/cpp/foxglove-websocket/CMakeLists.txt not work for you?

souryavarenya:
Hi @achim-k, unfortunately the existing CMakeLists.txt doesn't generate cmake package configuration. It only installs headers and libraries. For now, I have internally incorporated this repo as a submodule and wrapped it inside our custom CMakeLists.txt. You could check the closed PR diff to see the proposed change. The proposed change will work as a pure CMake build but needs more effort to make it work with Conan.

PR #560, opened 2023-10-23
Improve documentation of parameter types

achim-k:
### Public-Facing Changes

None

### Description
Addresses https://github.com/foxglove/ws-protocol/pull/519#issuecomment-1775641397


jtbandes (commented):
FWIW, I think "respectively" was used correctly, but this is fine too 😛

defunctzombie (approved):

defunctzombie:
> FWIW, I think "respectively" was used correctly, but this is fine too 😛

Yea - I found it required more mental energy to understand when reading the spec compared to enumerating them separately.

PR #563, opened 2023-10-25
spec: Allow specifying service schema names & encodings

achim-k:
### Public-Facing Changes

spec: Allow specifying service schema names & encodings

### Description
Allows for customization of service schema names and schema encodings.

Partially fixes https://github.com/foxglove/studio/issues/6970 
Partially resolves FG-5280

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -204,20 +204,20 @@ Informs the client about available services. Only supported if the server declar
 - `op`: string `"advertiseServices"`
 - `services`: array of:
   - `id`: number. The server may reuse ids when services disappear and reappear, but only if the services keeps the exact same name, type, and schema. Clients will use this unique id to cache schema info and deserialization routines.
-  - `name`: string
-  - `type`: string
-  - `requestSchema`: string, required if `request` is not given. Field is present for backwards compatibilty, prefer using `request` instead.
-  - `request`: object, required if `requestSchema` is not given.
-    - `encoding`: string
-    - `schemaName`: string
-    - `schemaEncoding`: string
-    - `schema`: string
-  - `responseSchema`: string, required if `response` is not given. Field is present for backwards compatibilty, prefer using `response` instead.
-  - `response`: object, required if `responseSchema` is not given.
-    - `encoding`: string
-    - `schemaName`: string
-    - `schemaEncoding`: string
-    - `schema`: string
+  - `name`: string, name used to identify the service.
+  - `type`: string, type of the service. May be used to derive request & response schema names if `request` or `response` are not given.
+  - `request`: object describing the request. Required if `requestSchema` is not given.
+    - `encoding`: string, type of encoding used for request message encoding.
+    - `schemaName`: string, name of the request schema.
+    - `schemaEncoding`: string, type of encoding used for schema encoding.
+    - `schema`: string, schema definition in a format matching the `schemaEncoding`.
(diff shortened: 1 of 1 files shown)


jhurliman:
Nice, LGTM

docs/spec.md
```diff
@@ -206,8 +206,14 @@ Informs the client about available services. Only supported if the server declar
   - `id`: number. The server may reuse ids when services disappear and reappear, but only if the services keeps the exact same name, type, and schema. Clients will use this unique id to cache schema info and deserialization routines.
   - `name`: string
   - `type`: string
-  - `requestSchema`: string
-  - `responseSchema`: string
+  - `requestSchema`: string (schema data only) or object with the fields below. If given as string, schema name and encoding will be derived from `type` (by appending `_Request`) and the previously advertised [supportedEncodings](#server-info).
```
jtbandes, line 209: I'm confused about what it means to derive the schema encoding from supportedEncodings, because I thought those were message encodings, not schema encodings?

jtbandes: I guess we also need a response message encoding in addition to response schema encoding?

achim-k: > I'm confused about what it means to derive the schema encoding from supportedEncodings, because I thought those were message encodings, not schema encodings?

As `serviceCallEncoding` (message encoding) we pick one of the supported encodings from  `supportedEncodings` and then derive the schema encoding from that:

https://github.com/foxglove/studio/blob/2c854ffb9c5a3676fb22818206d581bb1c08cf0f/packages/studio-base/src/players/FoxgloveWebSocketPlayer/index.ts#L606-L615

It's quite some magic. Maybe we just add `messageEncoding` to the objects that I have added in this PR?

jtbandes: Naming is hard... what do you think of:
```
request: {
  encoding: string
  schemaName: string
  schemaEncoding: string
  schema: string
}
```
or 
```
requestEncoding: string
requestSchema: {
  name: string
  encoding: string
  data: string
}
```

or something like that?

cc @defunctzombie for thoughts

achim-k: I am in favor of the 1st option

achim-k: Applied the changes. OK for you like that @jtbandes @jhurliman @defunctzombie ?

jtbandes (approved):
LGTM, would wait for @defunctzombie to sign off as well

defunctzombie (commented):
Maybe it is documented elsewhere or obvious to others but I have no idea what any of the fields in the `services` array are suppose to be. `id` has some more detailed documentation but all the other fields only have a plain type (string, object). Reading `request` I've no idea what I am suppose to think about this field, how do I populate it, what is it for? The same is true for `encoding`. How is that different than `schemaEncoding`? What would I typically put in these? How does it relate to the `schema` field?

All of those questions I cannot answer reading the spec. I see an example which helps but whether I would be able to turn that into an equivalent for other systems I am not sure.

Related: what is "encoding"? Why is it important to note `ros1` in the encoding? We already have a `schemaEncoding` of `ros1msg`. These are things that more descriptions on the fields might have made obvious but since there are not any I don't know what purpose this field serves that others have not already served.

defunctzombie (approved):

defunctzombie:
Awesome! Thank you for adding the descriptions. This is all clear to me now.

PR #579, opened 2023-11-06
@foxglove/ws-protocol: Update to spec changes

achim-k:
### Public-Facing Changes

@foxglove/ws-protocol: Update to spec changes

### Description
Adapts @foxglove/ws-protocol to the latest spec changes made in #563

Related to FG-5280

### DIFF
--- a/typescript/ws-protocol/src/types.ts
+++ b/typescript/ws-protocol/src/types.ts
@@ -40,12 +40,22 @@ export type Channel = {
   schema: string;
   schemaEncoding?: string;
 };
+
+type ServiceRequestDefinition = {
+  encoding: string;
+  schemaName: string;
+  schemaEncoding: string;
+  schema: string;
+};
+type ServiceResponseDefinition = ServiceRequestDefinition;
 export type Service = {
   id: number;
   name: string;
   type: string;
-  requestSchema: string;
-  responseSchema: string;
+  request?: ServiceRequestDefinition; // Must be given if requestSchema is not given.
+  response?: ServiceResponseDefinition; // Must be given if responseSchema is not given.
+  requestSchema?: string; // Must be given if request is not given.
+  responseSchema?: string; // Must be given if response is not given.
 };
 
 export type Subscribe = {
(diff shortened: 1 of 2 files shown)


typescript/ws-protocol/src/types.ts
```diff
@@ -40,13 +40,27 @@ export type Channel = {
...
+
+type ServiceRequest = {
+  encoding: string;
+  schemaName: string;
+  schemaEncoding: string;
+  schema: string;
+};
+type ServiceRequestDef =
+  | { requestSchema: string; request?: ServiceRequest }
+  | { requestSchema?: string; request: ServiceRequest };
```
jtbandes, line 52: I'm a bit confused by these types – is it allowed to specify both request & requestSchema? It seems like the types would allow that

achim-k: It's allowed, but only `request` would be used then. Better to make it exclusive?

typescript/ws-protocol/src/types.ts
```diff
@@ -40,12 +40,22 @@ export type Channel = {
...
 export type Service = {
   id: number;
   name: string;
   type: string;
-  requestSchema: string;
-  responseSchema: string;
+  request?: ServiceRequestDefinition; // Must be given if requestSchema is not given.
+  response?: ServiceResponseDefinition; // Must be given if responseSchema is not given.
+  requestSchema?: string; // Must be given if request is not given.
+  responseSchema?: string; // Must be given if response is not given.
```
jtbandes, line 58: Should we also mark these as deprecated, or maybe not since we will probably keep supporting it for a long time?

typescript/ws-protocol/src/types.ts
```diff
@@ -40,12 +40,22 @@ export type Channel = {
...
   schemaEncoding?: string;
 };
+
+type ServiceRequestDefinition = {
+  encoding: string;
+  schemaName: string;
+  schemaEncoding: string;
+  schema: string;
+};
+type ServiceResponseDefinition = ServiceRequestDefinition;
```
jtbandes, line 50: 🫠

jtbandes (approved):

PR #591, opened 2023-11-15
Add schemaEncoding to mcap-play channel registration

defunctzombie:
This conveys the schema encoding to the client allowing it to use the correct parser to handle the schema.


jtbandes (approved):

jtbandes:
Adding for context that I believe this is specifically necessary for IDL messages because the `cdr` message encoding can be used with either `ros2msg` or `ros2idl` schema encoding. If the client does not know which one to use it is forced to assume one based on `cdr` and it chooses `ros2msg` by default.

jtbandes:
We should also bump the examples package version and push the appropriate tag so it gets released to NPM

PR #598, opened 2023-11-22
mcap-record: Fix race condition when adding messages

### LINKED ISSUE #592: The data recorded by the mcap-record command is too large.
**Description**
When using the mcap-record command to record a live stream, if the recording file exceeds 700k (successfully recorded a 608k file), the mcap file will suddenly increase in size. When playing in foxglove, the error message "This file is unindexed. Unindexed files may have degraded performance" or "Chunks without message indexes are not currently supported" will be prompted.

- Version: loaded by npx,  it should be latest version I guess
- Platform: windows with node.js. npm version is  10.2.3, node version is  v21.2.0

**Steps To Reproduce**

the python example server could reproduce this. just play it long enough. The file suddenly takes up more space.

[mcap.zip](https://github.com/foxglove/ws-protocol/files/13372901/mcap.zip)
The attachment contains the results of two recordings, with only one frame difference between the two results.
**Expected Behavior**

achim-k:
### Public-Facing Changes

mcap-record: Fix race condition when adding messages

### Description
Does not directly call `writer.addMessage` in the websocket message handler but instead pushes it to a queue. This makes sure that we do not call `writer.addMessage` concurrently which could happen due to a certain `addMessage` taking longer when a chunk is ready to be compressed & written. This PR also allows to set the message queue size and to opt out of chunk compression.

Fixes #592
Resolves FG-5775


typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -190,13 +197,25 @@ async function main(address: string, options: { output: string }): Promise<void>
...
         }
-        void writer.addMessage({
-          channelId: subscription.mcapChannelId,
-          sequence: subscription.messageCount++,
-          logTime: BigInt(Date.now()) * 1_000_000n,
-          publishTime: event.timestamp,
-          data: new Uint8Array(event.data.buffer, event.data.byteOffset, event.data.byteLength),
-        });
+
+        // If we can't acquire the lock, we will drop the message to avoid potential OOM issues.
```
jtbandes, line 201: Maybe we could have a setting for this? It could be useful in some situations to record everything even if it takes longer than realtime. We should make sure that after ^C the writer still gets closed cleanly — ideally finishing everything that was in the write queue already

achim-k: done

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -98,13 +106,28 @@ async function main(address: string, options: { output: string }): Promise<void>
...
 
+  async function processMsgQueue() {
+    while (isProcessing) {
+      let message = messageQueue.shift();
+      while (message != undefined) {
+        await writer.addMessage(message);
+        message = messageQueue.shift();
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, 10));
```
jtbandes, line 120: Instead of a tight loop with a timeout, you could use `promise-queue`:

https://github.com/foxglove/mcap/blob/83c0428e93ae88df6f7360d9461d569be966fd0b/website/src/components/McapRecordingDemo/Recorder.ts#L47-L48

achim-k: thanks for the suggestion, just implemented it

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -190,13 +201,20 @@ async function main(address: string, options: { output: string }): Promise<void>
...
+            await writer.addMessage({
+              channelId: subscription.mcapChannelId,
+              sequence: subscription.messageCount++,
+              logTime: BigInt(Date.now()) * 1_000_000n,
+              publishTime: event.timestamp,
+              data: new Uint8Array(event.data.buffer, event.data.byteOffset, event.data.byteLength),
+            });
+          })
+          .catch(() => {
+            log(`Dropping message due to queue limit (${options.queueLimit}) reached.`);
```
jtbandes, line 216: would this also be triggered if the underlying addMessage resulted in an error? Shouldn't really matter in practice I guess

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -209,6 +227,13 @@ async function main(address: string, options: { output: string }): Promise<void>
         resolve();
       });
     });
+
+    const waitForWriteQueue = async () => {
+      while (writeMsgQueue.getPendingLength() + writeMsgQueue.getQueueLength() > 0) {
+        await new Promise((resolve) => setTimeout(resolve, 100));
+      }
+    };
+    await waitForWriteQueue();
```
jtbandes, line 236: I don't think you need the wrapper function here. You can just have the `while` loop.

achim-k:
@jtbandes ping

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -218,4 +243,11 @@ export default new Command("mcap-record")
   .description("connect to a WebSocket server and record an MCAP file")
   .argument("<address>", "WebSocket address, e.g. ws://localhost:8765")
   .option("-o, --output <file>", "path to write MCAP file")
+  .option("-n, --no-compression", "do not compress chunks")
+  .option(
+    "-q, --queue-limit <value>",
```
jtbandes, line 248: Is "queue limit" a common phrase? I think I've heard "queue size" more frequently?

jtbandes (approved):

typescript/ws-protocol-examples/src/examples/mcap-record.ts
achim-k: Looks like it, judging from the documentation of `PromiseQueue.add`. I think I will change it to just log the returned error here

> @returns
> A promise that forwards the resolution/rejection of the promise returned by promiseGenerator, or immediately rejects if maxQueuedPromise is exceeded.

PR #621, opened 2023-12-11
Add support for details in messages from the server

superyuuki:
### Public-Facing Changes

Adds an optional field to status messages from the server

### Description

Websocket servers can send status messages to the client where they will show up as problems, but they will always have a subtext of 'no details provided' because there is no way to supply details or 'tips' to the client despite the client having facilities to support it (in Problem) 

It would be nice to add a (maybe non-spec) optional field to the status message that will allow the client the possibility to display tips if the client supports tips, and otherwise be ignored

### DIFF
--- a/typescript/ws-protocol/src/types.ts
+++ b/typescript/ws-protocol/src/types.ts
@@ -140,7 +140,8 @@ export type ServerInfo = {
 export type StatusMessage = {
   op: "status";
   level: StatusLevel;
-  message?: string;
+  message: string;
+  tip?: string;
 };
 export type Advertise = {
   op: "advertise";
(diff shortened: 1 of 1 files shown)


superyuuki:
My understanding is that since the message is json anyways by the spec, the client can just load it into memory as a map and if there is no 'tips' provided then everything proceeds as-is

jtbandes:
Thanks for sharing the idea. I don't think we are interested in this feature right now since the user impact is fairly minimal, and usually an error coming from the websocket server would indicate some kind of system problem that should be resolved by making configuration changes, therefore most users of the app wouldn't be needing to see details about these errors. If there is some other motivating example, I didn't see it in your PR description (either this one or the foxglove/studio PR). There is also a simple workaround available because you can put the details in the `message` string itself. If we were to add this feature it would also need to be mentioned in the spec document and supported in all the different languages that we offer libraries for. I will close this PR since we don't plan to merge in its current form, but if you have more info about the use case to share feel free to discuss here.

PR #622, opened 2023-12-15
Document thread safety and message latching use case example

harveybia:
### Public-Facing Changes

No code changes. Updated documentation for c++ implementation of foxglove websocket.

### Description

- Document thread safety and message latching use case example


achim-k (approved):
Thank you for the contribution

PR #631, opened 2023-12-23
mcap-play: Play at wall speed

amacneil:
### Public-Facing Changes
`mcap-play` now attempts to play messages using `log_time` against the wall clock. Previously it was playing messages as fast as possible regardless of the `log_time` between messages.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 // -*- jsonc -*-
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,7 +1,7 @@
 // -*- jsonc -*-
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
defunctzombie, line 4:
```suggestion
    "source.fixAll.eslint": true
```

defunctzombie:
@jtbandes I've incorporated the _rate_ setting back into this PR. In my testing I was confused because the messages _are_ following the desired rate but since we don't use this setting to control the "clock" time in studio, it was still updating my fixed-window plots at wall time.

jtbandes (approved):
Variable names could use some clarity on what units they are in (e.g. `elapsedMessageTimeMs`)

PR #684, opened 2024-02-19
enable C++ compiler warnings

achim-k:
### Public-Facing Changes
None

### Description
- Enables compiler warnings
- Fixes warnings identified by @jhurliman (requires `-Winconsistent-missing-destructor-override` and `-Wextra-semi-stmt`)


jtbandes (approved):
I like how in order to fix the C++ lint warnings we also had to fix a TypeScript lint warning.

achim-k:
> I like how in order to fix the C++ lint warnings we also had to fix a TypeScript lint warning.

Oops, that change was accidentally committed. Will remove that commit again, was merged in #655 already. Wondering why Github does not mark it as conflicting changes :thinking:

PR #687, opened 2024-02-21
Improve cpp CMakeLists.txt

kyle-verdant:
### Public-Facing Changes

Removes pointless INSTALL step inside CMakeLists.txt, sets the library to be explicitly static.

### Description

Default for libraries in CMake is `STATIC`. Installation of static libraries is pretty pointless - if it's desired to key off of `BUILD_SHARED_LIBS` or similar, I recommend only installing if built as shared, use `CMAKE_LIST_DIR` to properly support `add_subdirectory`, and to rename `LICENSE` on install to `FOXGLOVE_LICENSE` or similar. This came up trying to use CPM to install (`FetchContent` would have similar issues)

### DIFF
--- a/cpp/foxglove-websocket/CMakeLists.txt
+++ b/cpp/foxglove-websocket/CMakeLists.txt
@@ -4,7 +4,7 @@ project(FoxgloveWebSocket CXX)
 find_package(nlohmann_json REQUIRED)
 find_package(websocketpp REQUIRED)
 
-add_library(foxglove_websocket src/base64.cpp src/parameter.cpp src/serialization.cpp src/server_factory.cpp)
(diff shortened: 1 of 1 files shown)


achim-k:
@kyle-verdant Some of the CI jobs are failing, could you look into these?

jlack1987:
I think we shouldn't remove the install lines no? That would break the cmake install target and will break the usage for ppl that are using this as a pure cmake project.

defunctzombie:
@kyle-verdant Are you still wanting to push this forward or has it gone stale and can be closed?

jtbandes:
Closing since this has been stale for over a month. If anyone is interested in pushing this forward we can reopen or create a new PR.

PR #725, opened 2024-04-06
Add provenance attestation when publishing to NPM

jtbandes:
### Changelog
None

### Description

This adds a provenance attestation to the published package so consumers can verify that the package was built on GitHub Actions:
- https://github.blog/2023-04-19-introducing-npm-package-provenance/
- https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions

The package will appear like this on npm:

Also fixes an [error when publishing](https://github.com/foxglove/ws-protocol/actions/runs/8197963504/job/22420730354) due to trying to publish the same package twice (with different node versions).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
       - name: Publish to NPM
+        # prevent trying to publish the same package twice with different node versions
         if: ${{ matrix.node-version == '18.x' && startsWith(github.ref, format('refs/tags/releases/typescript/{0}/v', matrix.package)) }}
         # `yarn publish` does not support --provenance
         run: npm publish typescript/${{ matrix.package }}/*.tgz --provenance --access public
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,16 +35,17 @@ jobs:
       - run: yarn workspace @foxglove/${{ matrix.package }} lint:ci
       - run: yarn workspace @foxglove/${{ matrix.package }} test
 
-      - run: yarn workspace @foxglove/${{ matrix.package }} pack
+      - if: ${{ matrix.package != 'test-client-web-app' }}
+        run: yarn workspace @foxglove/${{ matrix.package }} pack
       - name: Publish to NPM (dry run)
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -30,9 +35,17 @@ jobs:
...
+      - name: Publish to NPM (dry run)
+        if: ${{ matrix.node-version == '18.x' }}
+        # `yarn publish` does not support --provenance
+        run: npm publish typescript/${{ matrix.package }}/*.tgz --provenance --access public --dry-run
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
       - name: Publish to NPM
-        if: ${{ startsWith(github.ref, format('refs/tags/releases/typescript/{0}/v', matrix.package)) }}
-        run: yarn workspace @foxglove/${{ matrix.package }} publish --access public
+        if: ${{ matrix.node-version == '18.x' && startsWith(github.ref, format('refs/tags/releases/typescript/{0}/v', matrix.package)) }}
```
snosenzo, line 46: should there be a comment for this here?

.github/workflows/ci.yml
```diff
@@ -30,9 +35,17 @@ jobs:
       - run: yarn workspace @foxglove/${{ matrix.package }} lint:ci
       - run: yarn workspace @foxglove/${{ matrix.package }} test
 
+      - run: yarn workspace @foxglove/${{ matrix.package }} pack
+      - name: Publish to NPM (dry run)
+        if: ${{ matrix.node-version == '18.x' }}
+        # `yarn publish` does not support --provenance
```
snosenzo, line 41: you linked to the yarn berry issue in the mcap repo, would be good to have it here too:
https://github.com/yarnpkg/berry/issues/5430

jtbandes: This repo isn’t using berry for better or worse

defunctzombie: I think you can drop this yarn comment altogether. It isn't crazy that to publish to a repository you use the tool made by that repository.

defunctzombie (approved):

.github/workflows/ci.yml
jtbandes: I will keep the comment since I think the `yarn pack && npm publish package.tgz` is a bit unusual otherwise and I would wonder why it's being done this way.

PR #727, opened 2024-04-12
Fix onerror handler

eyr1n:
### Changelog
EventEmitter always emits values of `Error` types on "error", not `undefined`.

### Description

According to [the documentation of ws module for Node.js](https://github.com/websockets/ws/blob/HEAD/doc/ws.md#event-error), onerror handler emits `Error` types.
However, WebSocket in the browser does not emits `event.error`. This behavior is incorrect as per the code below.

https://github.com/foxglove/ws-protocol/blob/736c8dfc9256453a7ac017a823f308fd580571d2/typescript/ws-protocol/src/FoxgloveClient.ts#L29

I fixed to always emits values of `Error` types on "error".

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveClient.ts
+++ b/typescript/ws-protocol/src/FoxgloveClient.ts
@@ -73,8 +73,8 @@ export default class FoxgloveClient {
 
   #reconnect() {
     this.#ws.binaryType = "arraybuffer";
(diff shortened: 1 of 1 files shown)


achim-k (approved):
I think this is fine. AFAIU, in the browser the error event does not contain any information so just emitting a new Error seems OK

jtbandes (approved):

PR #733, opened 2024-04-23
Add operation `serviceCallFailure` to communicate failure of calling a service

achim-k:
### Changelog
Add operation `serviceCallFailure` to communicate failure of calling a service

### Description
The current spec lacks a way to communicate the failure of calling a service. This PR changes that by adding a new operation which serves that purpose.

PR #743, opened 2024-05-13
Add function to send service call failure to C++ implementation

achim-k:
### Changelog
Add function to send service call failure to C++ implementation

### Description
In #733 we have added an additional operation to communicate service call failures to the client. This PR adds the corresponding function `sendServiceFailure` to the C++ implementation.

### DIFF
--- a/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
@@ -149,6 +149,8 @@ public:
                            const std::string& message);
   void broadcastTime(uint64_t timestamp) override;
   void sendServiceResponse(ConnHandle clientHandle, const ServiceResponse& response) override;
+  void sendServiceFailure(ConnHandle clientHandle, ServiceId serviceId, uint32_t callId,
+                          const std::string& message) override;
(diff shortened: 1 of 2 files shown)


cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
```diff
@@ -149,6 +149,8 @@ class Server final : public ServerInterface<ConnHandle> {
                            const std::string& message);
   void broadcastTime(uint64_t timestamp) override;
   void sendServiceResponse(ConnHandle clientHandle, const ServiceResponse& response) override;
+  void sendServiceFailure(ConnHandle clientHandle, ServiceId serviceId, uint32_t callId,
```
jtbandes, line 152: nit: it is weird that the method name doesn't match the server op, e.g. "sendServiceCallFailure"

jtbandes (approved):

cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
achim-k: Agreed, but then I would also have to rename `sendServiceResponse` accordingly.

PR #744, opened 2024-05-14
mcap-record: Avoid duplicate subscription to channel

achim-k:
### Changelog
Fix `mcap-record` creating duplicated channels if existing channels are re-advertised by the server

### Docs
None

### Description
Logs a message when the server erroneously re-advertises channels which have been advertised before. As per the [spec](https://github.com/foxglove/ws-protocol/blob/main/docs/spec.md#advertise), only new channels should be advertised by the server but unfortunately not all servers comply with that. This extra check makes sure that channels are not duplicated in the resulting MCAP file.

### DIFF
--- a/typescript/ws-protocol-examples/src/examples/mcap-record.ts
+++ b/typescript/ws-protocol-examples/src/examples/mcap-record.ts
@@ -129,6 +129,7 @@ async function main(
         SubscriptionId,
         { messageCount: number; mcapChannelId: McapChannelId }
       >();
+      const knownChannelIds = new Set<WsChannelId>();
 
       client.on("serverInfo", (event) => {
         log(event);
@@ -143,6 +144,16 @@ async function main(
       client.on("advertise", (newChannels) => {
         void Promise.all(
           newChannels.map(async (channel) => {
+            if (knownChannelIds.has(channel.id as WsChannelId)) {
+              log(
+                "skipping channel %d on topic %s as a channel with the same id has been advertised before.",
+                channel.id,
+                channel.topic,
+              );
+              return;
+            }
+            knownChannelIds.add(channel.id as WsChannelId);
+
             let schemaEncoding = channel.schemaEncoding;
             if (schemaEncoding == undefined) {
               schemaEncoding = {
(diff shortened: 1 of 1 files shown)


typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -129,6 +129,7 @@ async function main(
         SubscriptionId,
         { messageCount: number; mcapChannelId: McapChannelId }
       >();
+      const knownChannelIds = new Set<WsChannelId>();
```
snosenzo, line 132: `activeChannelIds` would feel more appropriate, since we're clearing on `unadvertise`

snosenzo (approved):

typescript/ws-protocol-examples/src/examples/mcap-record.ts
achim-k:
```suggestion
      const activeChannelIds = new Set<WsChannelId>();
```

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -143,6 +144,16 @@ async function main(
       client.on("advertise", (newChannels) => {
         void Promise.all(
           newChannels.map(async (channel) => {
+            if (knownChannelIds.has(channel.id as WsChannelId)) {
```
achim-k, line 147:
```suggestion
            if (activeChannelIds.has(channel.id as WsChannelId)) {
```

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -143,6 +144,16 @@ async function main(
...
+              log(
+                "skipping channel %d on topic %s as a channel with the same id has been advertised before.",
+                channel.id,
+                channel.topic,
+              );
+              return;
+            }
+            knownChannelIds.add(channel.id as WsChannelId);
```
achim-k, line 155:
```suggestion
            activeChannelIds.add(channel.id as WsChannelId);
```

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -205,12 +216,20 @@ async function main(
...
             const subscriptionId = client.subscribe(channel.id) as SubscriptionId;
             subscriptionsById.set(subscriptionId, { messageCount: 0, mcapChannelId });
           }),
         );
       });
+
+      client.on("unadvertise", (channelIds) => {
+        for (const channelId of channelIds) {
+          log("channel %d has been unadvertised", channelId);
+          knownChannelIds.delete(channelId as WsChannelId);
```
achim-k, line 229:
```suggestion
          activeChannelIds.delete(channelId as WsChannelId);
```

typescript/ws-protocol-examples/src/examples/mcap-record.ts
```diff
@@ -205,12 +216,20 @@ async function main(
...
 
-            log("subscribing to %s", channel.topic);
+            log("subscribing to %s (channel %d)", channel.topic, channel.id);
             const subscriptionId = client.subscribe(channel.id) as SubscriptionId;
             subscriptionsById.set(subscriptionId, { messageCount: 0, mcapChannelId });
           }),
         );
       });
+
+      client.on("unadvertise", (channelIds) => {
```
jtbandes, line 226: this makes me wonder what happens if the websocket server unadvertises and then re-advertises the same channel id again? I forget if that is allowed in the protocol? If so I wonder if we would technically need to update `wsChannelsByMcapChannel` as well (or maybe that will cause more problems)

jtbandes (approved):

PR #776, opened 2024-07-08
Update eslint from 7.2.0 to 7.15.0

achim-k:
### Changelog
None

### Docs
None

### Description
Update eslint from 7.2.0 to 7.15.0


achim-k:
CI fails for Node 16 :(
```
The engine "node" is incompatible with this module. Expected version "^18.18.0 || >=20.0.0". Got "16.20.2"
```

Would it be OK to remove support for node < 18.18? Or we close this PR again

snosenzo:
@achim-k I am likely not qualified to answer that question 😅 Might need to wait til @defunctzombie returns

defunctzombie:
Node v16 is EOL: https://github.com/forcedotcom/cli/issues/1985

I think it is fair to drop ongoing releases for it.

We originally had it and added v18 via https://github.com/foxglove/ws-protocol/commit/d2b944a297c8175ed38fd95c5db12c6d06ccff4e

I think we should drop v16 and add the newer node versions in its place. We can stick to LTS + latest version(s) if we prefer to simplify this list.

defunctzombie (approved):

PR #783, opened 2024-07-17
Add support for clearing previously sent status messages

achim-k:
### Changelog
Add support for clearing previously sent status messages

### Docs
None

### Description
Adds the `removeStatus` operation which can be used to clear a previously sent status message. To identify which status message shall be cleared, a `id` field has been added to the status message.

Adapted implementations:
- [x] typescript
- [x] c++
- [x] python

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -25,7 +25,7 @@
 
 - [Server Info](#server-info) (json)
 - [Status](#status) (json)
-- [Clear Status](#clear-status) (json)
+- [Remove Status](#remove-status) (json)
 - [Advertise](#advertise) (json)
 - [Unadvertise](#unadvertise) (json)
 - [Message Data](#message-data) (binary)
@@ -102,7 +102,7 @@ Each JSON message must be an object containing a field called `op` which identif
 - `op`: string `"status"`
 - `level`: 0 (info), 1 (warning), 2 (error)
 - `message`: string
-- `id`: string | undefined. Optional identifier for the status message. Newer status messages with the same identifier should replace previous messages. [clearStatus](#clear-status) can reference the identifier to indicate a status message is no longer valid.
+- `id`: string | undefined. Optional identifier for the status message. Newer status messages with the same identifier should replace previous messages. [removeStatus](#remove-status) can reference the identifier to indicate a status message is no longer valid.
 
 #### Example
 
@@ -115,20 +115,20 @@ Each JSON message must be an object containing a field called `op` which identif
 }
 ```
 
-### Clear Status
+### Remove Status
 
 - Informs the client that previously sent status message(s) are no longer valid.
 
 #### Fields
 
-- `op`: string `"clearStatus"`
-- `statusIds`: array of string, ids of the status messages to be cleared. The array must not be empty.
+- `op`: string `"removeStatus"`
+- `statusIds`: array of string, ids of the status messages to be removed. The array must not be empty.
 
 #### Example
 
 ```json
 {
-  "op": "clearStatus",
+  "op": "removeStatus",
   "statusIds": ["status-123"]
 }
 ```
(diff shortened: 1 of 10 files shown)


snosenzo (commented):
More general question: should we add another method to clear all status messages? It feels like that might be helpful, though I know the ticket isn't explicitly asking for that

achim-k:
> More general question: should we add another method to clear all status messages? It feels like that might be helpful, though I know the ticket isn't explicitly asking for that

This might be helpful but I would stick with keeping the spec lean and only adding "atomic" operations. All higher-level methods have to be built up on these.

docs/spec.md
```diff
@@ -101,14 +102,34 @@ Each JSON message must be an object containing a field called `op` which identif
...
+```
+
+### Clear Status
+
+- Informs the client that a previously sent status message shall be cleared.
+
+#### Fields
+
+- `op`: string `"clearStatus"`
+- `id`: string, the id of the status message to be cleared.
```
achim-k, line 125: @snosenzo's [comment](https://github.com/foxglove/ws-protocol/pull/783?notification_referrer_id=NT_kwDOAI0la7MxMTU0NTY5MzU2MTo5MjUwMTU1&notifications_query=is%3Aunread#pullrequestreview-2183977854) got me thinking that this could also be a string array to allow clearing of multiple status at once

```suggestion
- `id`: string[], ids of the status messages to be cleared.
```

snosenzo: yeah I think that's a good improvement

docs/spec.md
```diff
@@ -101,14 +102,34 @@ Each JSON message must be an object containing a field called `op` which identif
 - `op`: string `"status"`
 - `level`: 0 (info), 1 (warning), 2 (error)
 - `message`: string
+- `id`: string | undefined. Optional unique identifier that can be used for clearing the status.
```
defunctzombie, line 105: nit: "Optional identifier for the status message. Newer status messages with the same identifier should replace previous messages. `clearStatus` can reference the identifier to indicate a status message is no longer valid."

defunctzombie (commented):
Idea: re-purpose the status field and treat an empty message (with an `id` provided) as "clearing" the status.

defunctzombie (approved):
This seems fine - you could avoid the extra op by introducing some new semantics to the `status` OP (since an empty status message is a decent sentinel) but if you like the more explicit nature of "clearStatus" that seems ok too.

achim-k:
> This seems fine - you could avoid the extra op by introducing some new semantics to the `status` OP (since an empty status message is a decent sentinel) but if you like the more explicit nature of "clearStatus" that seems ok too.

I prefer having a dedicated operation here. Sending an empty status message would definitely work, but we would have to also mention that the `level` does not matter then which in summary seems a little confusing.

docs/spec.md
```diff
@@ -101,14 +102,34 @@ Each JSON message must be an object containing a field called `op` which identif
 - `op`: string `"status"`
 - `level`: 0 (info), 1 (warning), 2 (error)
 - `message`: string
+- `id`: string | undefined. Optional identifier for the status message. Newer status messages with the same identifier should replace previous messages. [clearStatus](#clear-status) can reference the identifier to indicate a status message is no longer valid.
```
snosenzo, line 105: seems odd to me that the ID is optional here but not optional for `clearStatus`

snosenzo (approved):
Looks good to me, but I do think it's odd that we allow sending status messages without Ids, but don't allow clearing of all status by passing an empty array or making the statusIds array option

snosenzo:
Only other thought that just occurred, was that maybe this method should be called `removeStatus` instead of `clearStatus` so that we can use the latter for allowing to clear status without specifying id's?

docs/spec.md
achim-k: The `id` field has to be optional here to keep the `status` operation backwards compatible.

snosenzo (approved):

snosenzo:
only thing is don't forget to update PR description from `clearStatus` to `removeStatus`

PR #817, opened 2024-09-07
Add option to build as an ament package

jlack1987:
### Changelog
Add a `package.xml` and update the `CMakeLists.txt` to optionally allow users to build as an ament package.

### Docs

None

### Description

I added a `package.xml` and some updates to the `CMakeLists.txt` so this package could be used in ros2 workspaces. Note that this should only have an affect if it is already in a ros2 workspace and the `ament_cmake` package is found. If not then it should continue to compile as before.

### Help Needed
I wasn't sure what to put for some of the fields in the `package.xml`, so if this is acceptable feel free to edit and modify maintainer info or you can tell me what it should be and i'll add it. Thanks!

### DIFF
--- a/cpp/foxglove-websocket/CMakeLists.txt
+++ b/cpp/foxglove-websocket/CMakeLists.txt
@@ -1,13 +1,21 @@
 cmake_minimum_required(VERSION 3.15)
 project(FoxgloveWebSocket CXX)
 
+find_package(ament_cmake QUIET)
+set(COMPILING_FOR_ROS2 OFF)
+if(${ament_cmake_FOUND})
+  set(COMPILING_FOR_ROS2 ON)
+  message(STATUS "Compiling for ROS2")
+endif()
+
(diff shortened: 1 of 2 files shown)


jlack1987:
actually this maybe isn't good. I see there are an examples package which I could make an ament packages as well, but it depends on foxglove protobuf schema stuff which i'm not really wanting to ament-ize all of that. Also it seems the foxglove_rosbridge package has a parallel implementation of the websocket protocol and seems to be more up to date? :confused:

jtbandes:
Closing since this has been stale for over a month. If anyone is interested in pushing this forward we can reopen or create a new PR.

> Also it seems the foxglove_rosbridge package has a parallel implementation of the websocket protocol and seems to be more up to date? 😕

We do try to keep them relatively in sync, although it may not be perfectly in sync at all times these days – cc @achim-k

PR #850, opened 2024-10-22
call action handlers after action, in c++ server

paulsohn:
### Changelog

Guarantees that action handlers in C++ ws server are called after the action has been applied.

### Docs

https://github.com/orgs/foxglove/discussions/1049

#### Problem

In C++ server, handlers for below events are called before the internal states are fully updated:
* subscribe
* unsubscribe
* client advertise
* client unadvertise
* subscribe connection graph

As a consequence, some user behavior does not work as expected, such as calling `sendMessage()` for messages of the requested topic inside of subscription handler. When subscription handler is called, the internal state does not recognize the requested topic as subscribed, and `sendMessage()` filters out this unsubscribed topic.

This seems to be a bug rather than a spec, because typescript and python server implementations calls listener/emit event after all internal tables are updated.

#### Changes

For the above events, switch the order of handler call and internal map updates.

https://linear.app/foxglove/issue/FG-9181/c-foxglove-websocket-server-ambiguous-handler-invoke-timing

### DIFF
--- a/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
@@ -1267,9 +1267,11 @@ void Server<ServerConfiguration>::handleSubscribe(const nlohmann::json& payload,
       continue;
     }
 
+    {
+      std::unique_lock<std::shared_mutex> clientsLock(_clientsMutex);
+      _clients.at(hdl).subscriptionsByChannel.emplace(channelId, subId);
+    }
     _handlers.subscribeHandler(channelId, hdl);
-    std::unique_lock<std::shared_mutex> clientsLock(_clientsMutex);
-    _clients.at(hdl).subscriptionsByChannel.emplace(channelId, subId);
   }
 }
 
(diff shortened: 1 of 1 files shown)


cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
```diff
@@ -1336,10 +1340,12 @@ void Server<ServerConfiguration>::handleAdvertise(const nlohmann::json& payload,
     advertisement.encoding = chan.at("encoding").get<std::string>();
     advertisement.schemaName = chan.at("schemaName").get<std::string>();
 
+    {
+      std::unique_lock<std::shared_mutex> clientsLock(_clientsMutex);
+      _clients.at(hdl).advertisedChannels.emplace(channelId);
+      clientPublications.emplace(channelId, advertisement);
```
jtbandes, line 1346: I don't think this one needs to be under the lock, it's a local variable

jtbandes (approved):

achim-k (approved):

cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
```diff
@@ -1336,10 +1340,12 @@ void Server<ServerConfiguration>::handleAdvertise(const nlohmann::json& payload,
...
+    }
```
jtbandes, line 1347:
```suggestion
    clientPublications.emplace(channelId, advertisement);
    {
      std::unique_lock<std::shared_mutex> clientsLock(_clientsMutex);
      _clients.at(hdl).advertisedChannels.emplace(channelId);
    }
```

jtbandes:
Thanks!

PR #859, opened 2024-10-30
Add workflow to approve dependabot PRs

defunctzombie:
### Changelog
None

### Docs
None

### Description
Auto-approve and merge dependabot PRs.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,30 @@
+# Automatically approve dependabot PRs.
+
+name: Dependabot
+
+on:
+  - pull_request_target
+
+jobs:
+  approve:
+    name: Approve Dependabot PRs
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+
+    # IMPORTANT: setting YARN_ENABLE_SCRIPTS=false is critical to ensure that untrusted
+    # PRs can't add an npm package and then use that to execute untrusted code in
+    # a trusted context. See links at the top of this workflow for further details.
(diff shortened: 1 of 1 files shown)


defunctzombie:
@jtbandes @achim-k the linear issue asks for auto-merge as well. I can add that if we think it would be safe and appropriate?

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  approve:
+    name: Approve Dependabot PRs
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+
+    # IMPORTANT: setting YARN_ENABLE_SCRIPTS=false is critical to ensure that untrusted
+    # PRs can't add an npm package and then use that to execute untrusted code in
+    # a trusted context. See links at the top of this workflow for further details.
```
achim-k, line 16: Outdated commented? There are no links at the top of this workflow

achim-k (approved):

achim-k:
Auto-merge would be useful. I believe that we have enough tests in place to make it safe enough

.github/workflows/dependabot.yml
defunctzombie: Yea - the links part is outdated but it seemed nice to keep this env var as a precaution? Though this PR doesn't run yarn anymore now that I think about it

jtbandes (approved):
Seems alright. Will point out that we don't seem to have any tests for FoxgloveClient right now.

defunctzombie:
> Will point out that we don't seem to have any tests for FoxgloveClient right now.

Perfect 👏

PR #869, opened 2024-10-31
Feature/ament

jlack1987:
### Changelog
Add option to build as part of a ros2 workspace

### Description

Add optional ament check so ros2 users can compile this within a ros2 workspace and depend on it that way.


achim-k:
For ROS2 users, I would recommend to create a separate `ws_protocol_vendor` package (in another repo) which exposes the C++ & Python libs to ROS2 packages that depend on it.  This would look very similar to [mcap_vendor](https://github.com/ros2/rosbag2/blob/humble/mcap_vendor/CMakeLists.txt) which uses [fetchContent](https://cmake.org/cmake/help/v3.14/module/FetchContent.html) under the hood. We could then also use the `ws_protocol_vendor` package for https://github.com/foxglove/ros-foxglove-bridge/issues/89

Issue #875, opened 2024-11-12
C++ type definition doesn't contain schemaEncoding field


paulsohn:
**Description**

https://github.com/foxglove/ws-protocol/blob/a28e0b9c9c319a1803bc00184d76fdb97714e921/cpp/foxglove-websocket/include/foxglove/websocket/common.hpp#L77-L83

Tried to use client publish capability, but currently C++ type definition doesn't contain schemaEncoding field.

- Version: main
- Platform:

**Steps To Reproduce**

N/A

**Expected Behavior**
should contain schemaEncoding in the definition, and adjust server behavior if needed

paulsohn:
Sorry for misplaced ticket title.

jtbandes:
Additionally it seems `schema` is never populated when received by the server... https://github.com/foxglove/ws-protocol/blob/a28e0b9c9c319a1803bc00184d76fdb97714e921/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp#L1337-L1341

And is not included in the message sent from the client: https://github.com/foxglove/ws-protocol/blob/a28e0b9c9c319a1803bc00184d76fdb97714e921/cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp#L22-L25

paulsohn:
@jtbandes 

If it's just unimplemented feature, then `schema` might take some breaking change as well like
```cpp
struct ClientAdvertisement {
  // ...omit
  std::string schemaName;
  std::optional<std::vector<uint8_t>> schema; // or even std::optional<std::string> schema;
  std::optional<std::string> schemaEncoding;
};
```

PR #876, opened 2024-11-13
C++: add missing implementations for client advertised schema/encoding

paulsohn:
### Changelog

C++ server/client: add missing implementations for client advertised schema / schema encoding

### Docs

* Issue #875 (Initial discussions are here)
* Liner Ticket FG-9447 (though I can't access to it)

### Description

Although foxglove ws protocol allow client to advertise complete schema definition string as well as schema name, current C++ implementation lacks such fields. Particularly:
* `ClientAdvertisement` struct lacks `schemaEncoding` field,
* `ClientAdvertisement` struct has `schema` field, but it is never populated (websocket server does not read schema or schema encoding to json)
* websocket client is does not serialize schema or schema encoding to json

This PR resolves above problems, and additionally since this is breaking change from unimplemented to implemented, types are decided as following:
* `schema` field type is changed from `std::vector<uint8_t>` to `std::optional<std::string>`, as schema fields are passed in as json quoted string, and it's optional
* `schemaEncoding` field type is determined as `std::optional<std::string>`, the same reason for `schema`.

Fixes: foxglove/ws-protocol#875
Fixes: FG-9447

### DIFF
--- a/cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp
@@ -23,6 +23,12 @@ inline void to_json(nlohmann::json& j, const ClientAdvertisement& p) {
                      {"topic", p.topic},
                      {"encoding", p.encoding},
                      {"schemaName", p.schemaName}};
+  if (p.schema) {
+    j["schema"] = *p.schema;
+  }
+  if (p.schemaEncoding) {
+    j["schemaEncoding"] = *p.schemaEncoding;
+  }
 }
 
 using TextMessageHandler = std::function<void(const std::string&)>;
--- a/cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp
@@ -24,10 +24,10 @@ inline void to_json(nlohmann::json& j, const ClientAdvertisement& p) {
                      {"encoding", p.encoding},
                      {"schemaName", p.schemaName}};
   if (p.schema) {
-    j["schema"] = *p.schema;
+    j["schema"] = p.schema.value();
   }
   if (p.schemaEncoding) {
-    j["schemaEncoding"] = *p.schemaEncoding;
+    j["schemaEncoding"] = p.schemaEncoding.value();
   }
 }
 
--- a/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
+++ b/cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
@@ -1340,6 +1340,14 @@ void Server<ServerConfiguration>::handleAdvertise(const nlohmann::json& payload,
     advertisement.encoding = chan.at("encoding").get<std::string>();
     advertisement.schemaName = chan.at("schemaName").get<std::string>();
 
+    advertisement.schema = chan.find("schema") == chan.end()
+                             ? std::nullopt
+                             : std::optional<std::string>(chan["schema"].get<std::string>());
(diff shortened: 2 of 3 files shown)


defunctzombie:
@paulsohn Thanks for the proposed fix here - let us know when this is ready for review!

paulsohn:
@defunctzombie Sure.
It is a draft because I got no tests but I don't think further changes might be needed, so feel free to test this as it is (I will do my tests in next week)

paulsohn:
Fixed error.

Added some test in separate branch ( https://github.com/paulsohn/ws-protocol/tree/test-pr876 ).
After build, execute `make example_server_client_pub_json` and `make example_client_pub_json` on separate shell in this order.

paulsohn:
Are you currently reviewing this? If you don't mind, let me know the progress so far

By the way I just thought, it might affect ros-foxglove-bridge as well?

https://github.com/foxglove/ros-foxglove-bridge/blob/main/ros2_foxglove_bridge/src/ros2_foxglove_bridge.cpp

cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
```diff
@@ -1340,6 +1340,14 @@ void Server<ServerConfiguration>::handleAdvertise(const nlohmann::json& payload,
...
     advertisement.schemaName = chan.at("schemaName").get<std::string>();
 
+    advertisement.schema = chan.find("schema") == chan.end()
+                             ? std::nullopt
+                             : std::optional<std::string>(chan["schema"].get<std::string>());
+    advertisement.schemaEncoding =
+      chan.find("schemaEncoding") == chan.end()
+        ? std::nullopt
+        : std::optional<std::string>(chan["schemaEncoding"].get<std::string>());
+
```
achim-k, line 1350: I believe the following should also work which avoids a second lookup of the json property

```suggestion
    if (const auto it = chan.find("schema"); it != chan.end()) {
      advertisement.schema = it->get<std::string>();
    }
    if (const auto it = chan.find("schemaEncoding"); it != chan.end()) {
      advertisement.schemaEncoding = it->get<std::string>();
    }
```

cpp/foxglove-websocket/include/foxglove/websocket/websocket_client.hpp
```diff
@@ -23,6 +23,12 @@ inline void to_json(nlohmann::json& j, const ClientAdvertisement& p) {
                      {"topic", p.topic},
                      {"encoding", p.encoding},
                      {"schemaName", p.schemaName}};
+  if (p.schema) {
+    j["schema"] = *p.schema;
+  }
+  if (p.schemaEncoding) {
+    j["schemaEncoding"] = *p.schemaEncoding;
+  }
```
achim-k, line 31:
```suggestion
  if (p.schema) {
    j["schema"] = p.schema.value();
  }
  if (p.schemaEncoding) {
    j["schemaEncoding"] = p.schemaEncoding.value();
  }
```

To be consistent with the rest of the codebase

achim-k (approved):

cpp/foxglove-websocket/include/foxglove/websocket/websocket_server.hpp
paulsohn: Some of the previous code accessed fields like this and others used iterators just as you're suggesting, so I just picked one randomly. Same for the another suggestion.

Nonetheless, I agree with both of your points. Should I make a new commit or will you?

achim-k: Ah ok, was not aware of this.

jtbandes (approved):

jtbandes:
Thank you!

PR #916, opened 2025-01-30
[typescript] Allow client to understand other subprotocols

bryfox:
### Changelog

None

### Docs

None

### Description

This removes the "supported subprotocol" check performed after the handshake, instead relying on the underlying websocket client to advertise the correct subprotocol.

The `FoxgloveClient` already accepts a websocket client (`IWebSocket`), which is free to choose its advertised subprotocols. This change would allow it to be extended with backwards-compatible subprotocols (for example, v2).

This change is not breaking as long as existing clients are advertising the correct subprotocol.

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveClient.ts
+++ b/typescript/ws-protocol/src/FoxgloveClient.ts
@@ -58,6 +58,8 @@ const textEncoder = new TextEncoder();
  * "foxglove.websocket.v1").
  */
 export default class FoxgloveClient {
+  static SUPPORTED_SUBPROTOCOL = "foxglove.websocket.v1";
+
   #emitter = new EventEmitter<EventTypes>();
   #ws: IWebSocket;
   #nextSubscriptionId = 0;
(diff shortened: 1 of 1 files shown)


jtbandes (commented):
Hm, maybe we should just delete this check? I don’t think it’s really necessary anyway as the underlying WebSocket will handle the protocol as part of the handshake

bryfox:
> Hm, maybe we should just delete this check? I don’t think it’s really necessary anyway as the underlying WebSocket will handle the protocol as part of the handshake

I think that's reasonable. I presume this is a check to make sure the provided ws client is advertising a subprotocol that this library understands (and so can provide a better error), but we could document instead.

jtbandes (approved):

typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -49,6 +49,14 @@ type EventTypes = {
 
 const textEncoder = new TextEncoder();
 
+/**
+ * A client to make it easier to interact with the Foxglove WebSocket protocol:
```
jtbandes, line 53:
```suggestion
 * A client to interact with the Foxglove WebSocket protocol:
```

PR #926, opened 2025-03-12
Typescript: Use ArrayBufferView instead of DataView for data types.

achim-k:
### Changelog
Typescript: Use ArrayBufferView instead of DataView for data types.

### Docs
None

### Description
- 65323c5dcc6dc8228bcfc55485984ef162d927e1 changes the typescript client's `onmessage` handler to accept messages of type `ArrayBufferView` (e.g. Uint8Arrays). This gives more flexibility when implementing a custom `IWebSocket` source.
- 6382cea5da3f40ad98c898a651bbd3b28b39793b changes the type of `data` message fields from `DataView` to `ArrayBufferView`. This is a breaking change, but makes the API easier to use.

### DIFF
--- a/typescript/ws-protocol/src/FoxgloveClient.ts
+++ b/typescript/ws-protocol/src/FoxgloveClient.ts
@@ -91,10 +91,10 @@ export default class FoxgloveClient {
     this.#ws.onopen = (_event) => {
       this.#emitter.emit("open");
     };
-    this.#ws.onmessage = (event: MessageEvent<ArrayBuffer | string>) => {
+    this.#ws.onmessage = (event: MessageEvent<ArrayBuffer | ArrayBufferView | string>) => {
       let message: ServerMessage;
       try {
-        if (event.data instanceof ArrayBuffer) {
+        if (event.data instanceof ArrayBuffer || ArrayBuffer.isView(event.data)) {
           message = parseServerMessage(event.data);
         } else {
           message = JSON.parse(event.data) as ServerMessage;
--- a/typescript/ws-protocol/src/FoxgloveServer.test.ts
+++ b/typescript/ws-protocol/src/FoxgloveServer.test.ts
@@ -4,7 +4,6 @@ import { AddressInfo, Data, WebSocket, WebSocketServer } from "ws";
 import {
   BinaryOpcode,
   ClientBinaryOpcode,
-  ClientPublish,
   FetchAssetResponse,
   FetchAssetStatus,
   IWebSocket,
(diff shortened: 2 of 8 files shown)


typescript/ws-protocol/src/FoxgloveClient.ts
```diff
@@ -91,10 +91,10 @@ export default class FoxgloveClient {
     this.#ws.onopen = (_event) => {
       this.#emitter.emit("open");
     };
-    this.#ws.onmessage = (event: MessageEvent<ArrayBuffer | string>) => {
+    this.#ws.onmessage = (event: MessageEvent<ArrayBuffer | ArrayBufferView | string>) => {
       let message: ServerMessage;
       try {
-        if (event.data instanceof ArrayBuffer) {
+        if (event.data instanceof ArrayBuffer || ArrayBuffer.isView(event.data)) {
           message = parseServerMessage(event.data);
```
jtbandes, line 98: I think this might be wrong, if it is an ArrayBufferView, then we need to respect the `byteOffset` and `byteLength`

typescript/ws-protocol/src/FoxgloveServer.test.ts
```diff
@@ -280,14 +279,9 @@ describe("FoxgloveServer", () => {
...
           channel: { id: 42, topic: "foo", encoding: "json", schemaName: "baz" },
-          data: new DataView(expectedPayload.buffer),
+          data: expectedPayload,
         },
       ]);
-      const msg = msgEvent[1] as ClientPublish;
-      expect(msg.data.byteLength).toEqual(expectedPayload.byteLength);
-      for (let i = 0; i < expectedPayload.byteLength; i++) {
-        expect(msg.data.getUint8(i)).toEqual(expectedPayload[i]);
-      }
```
jtbandes, line 290: wonder why this was here? was jest not able to compare DataView by default?

jtbandes (approved):
curious what's the motivation for this? is this so it can be used with SharedArrayBuffer?

minor nit: I wonder if the extra accesses to `view.buffer`, `view.byteOffset` etc. might be a slight performance degradation if they are not optimized out, could consider hoisting these variables. But on the other hand we are already doing `new DataView()` for every message so it's probably not a big deal

typescript/ws-protocol/src/FoxgloveClient.ts
jtbandes: Do we need to make the corresponding change here?


https://github.com/foxglove/ws-protocol/blob/6382cea5da3f40ad98c898a651bbd3b28b39793b/typescript/ws-protocol/src/FoxgloveServer.ts#L391-L392

achim-k: yes, good catch

typescript/ws-protocol/src/FoxgloveServer.test.ts
achim-k: It was able to compare DataView. No idea why it was here in the first place

PR #927, opened 2025-03-13
Bump `@foxglove/ws-protocol` to 0.8.0, `@foxglove/ws-protocol-examples` to 0.10.0

achim-k:
### Changelog
#926

### DIFF
--- a/test-client-web-app/package.json
+++ b/test-client-web-app/package.json
@@ -7,7 +7,7 @@
     "@emotion/react": "^11.13.5",
     "@emotion/styled": "^11.13.0",
     "@fontsource/roboto": "^5.1.0",
-    "@foxglove/ws-protocol": "^0.7.1",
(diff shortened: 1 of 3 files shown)


test-client-web-app/package.json
```diff
@@ -7,7 +7,7 @@
     "@emotion/react": "^11.13.5",
     "@emotion/styled": "^11.13.0",
     "@fontsource/roboto": "^5.1.0",
-    "@foxglove/ws-protocol": "^0.7.1",
+    "@foxglove/ws-protocol": "0.8.0",
```
achim-k, line 10: should this be a major version bump instead?

jtbandes: 🤷  I think when the major version is 0, then it's allowed to have breaking changes that increment the minor version

jtbandes (approved):

